PR #116, opened 2017-09-05
replaced StateMap and Stylesheet interface declaration with type declaration

wtfil:
Hi. 
I am trying to create stylable stateless component

```tsx
import * as React from 'react';
import {SBStateless} from 'stylable-react-component';
import styles from './styles.st.css';

export const MyComponent = SBStateless(
    (props) => {
        return (...);
    },
    styles
);
```

But I receive error like

`error TS4023: Exported variable 'MyComponent' has or is using name 'Stylesheet' from external module "...PATH.../node_modules/stylable/dist/src/runtime" but cannot be named.`

I made some research and found this [issue](https://github.com/Microsoft/TypeScript/issues/5711), which said that I either have to add explicit import of `Stylesheet` in my component or let compiler add type declaration automatically.
I prefer the second option but this require the update type declaration. As far I know, `type` and `interface` has only different that `type` will not create actual alias name, so compiler could use this.

### DIFF
--- a/src/runtime.ts
+++ b/src/runtime.ts
@@ -1,7 +1,7 @@
 
-export interface StateMap { [key: string]: boolean }
+export type StateMap = { [key: string]: boolean }
 
-export interface Stylesheet {
+export type Stylesheet = {
     namespace: string;
     root: string;
     get: (localName: string) => string;


AviVahl:
Does the validation error occur if you give MyComponent an explicit type?
for example:
https://github.com/wix/stylable-components/blob/master/src/components/drop-down/drop-down.tsx#L29

AviVahl:
Also, TS 2.6 is expected to address it, as far as I can see (was added to the milestone).
The compiler will add the imports to the .d.ts automatically.
https://github.com/Microsoft/TypeScript/issues/9944

wtfil:
@AviVahl no, it does not. Thx!
I think i can use that for now, and probably this PR is not relevant

PR #120, opened 2017-09-07
add react integration docs

idoros:


docs/guides/react-integration.md
```diff
@@ -1,8 +1,71 @@
...
+```
+
+Using yarn:
+```bash
+yarn add wix-react-tools
+```
+
+## Use
+
+When applying Stylable to a React component, any className or `data-*` properties are copied to the resulting root element of the component. Further more the [root class](../references/root.md) of the stylesheet is added to the root element automatically.
```
brucelawson: "Furthermore" is one word

PR #134, opened 2017-09-12
custom-selectors

idoros:
[Custom selectors](https://drafts.csswg.org/css-extensions/#custom-selectors) basic polyfill with exposure as pseudo-elements


site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+> **Notice**:  
+> In case a `custom-selector` alias conflicts with a local CSS class name, the exposed `pseudo-element` will target the `custom-selector`. However the the exported CSS class will still be exported to Javascript.
+
+## Use cases
+
+### Container and recursive components
+
+Some components might contain nested instances of themselves, because they're a container or a "recursive" component (e.g. Tree component might render itself). 
+
+In case the component exposes any `pseudo-elements`, it is a good practice to define them as `custom-selectors` with [child selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_selectors) to avoid effecting internal instances inner parts.
```
arnonkehat: affecting

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+@namespace "Panel";
+:import {
+    -st-from: "./tree.st.css";
+    -st-default: Tree;
+}
+/*
+selector: .Panel__root .Tree__root > .Tree__icon
+*/
+Tree::icon {
+    background: yellow;
```
arnonkehat: need an explanation of the outcome. like /* will paint icons all the way down the tree */

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+:import {
+    -st-from: "./gallery.st.css";
+    -st-default: Gallery;
+}
+@custom-selector :--navigationBtn Gallery::navBtn;
+```
+
+### Combination selector
+
+Sometimes a component has several basic CSS classes (with corresponding `pseudo-elements`) and we want to expose a combination `pseudo-element`.
```
arnonkehat: i don't understand this premise? are we just connecting two classes with a single selector?

arnonkehat:
i think you can safely merge this with a big warning at the top that this is an experimental feature, and the doc is still a draft.

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+```css
+@namespace "Page";
+:import {
+    -st-from: "./comp.st.css";
+    -st-default: Comp;
+}
+/*
+selector: .Page__root .Comp__root .Comp__controls .Comp__btn
+*/
+Comp::controlBtn {
```
brucelawson: Should we adhere to the (proto-)spec and use the empty vendor prefix at the beginning of a custom selectorl, eg Comp::--controlBtn {color: goldenrod;} ?

I suggest we should, to keep it closer to spec.

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+```css
+@namespace "Comp";
+@custom-selector :--symbol .icon, .thumb, .picture;
+/*
+selector: 
+.Comp__root .Comp__icon, 
+.Comp__root .Comp__thumb, 
+.Comp__root .Comp__picture
+*/
+:--heading {
```
brucelawson: where is :--heading defined? (Is it meant to be :--symbol, defined in line 121?)

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,164 @@
...
+.Comp__root .Comp__picture
+*/
+:--heading { 
+    border: 1px solid grey; 
+}
+```
+
+#### Caveats
+
+Aliasing multiple selectors in a `custom-selector` is a [footgun feature](https://en.wiktionary.org/wiki/footgun#English) that might generate lots of CSS.
```
brucelawson: Should the user care about this "footgun feature"? Why should they care that lots of CSS is generated? File size for text is trivial

site/docs/references/custom-selectors.md
idoros: Agreed

site/docs/references/custom-selectors.md
idoros: Where in the spec does custom selectors used as pseudo-elements? I like to take the base name and not the full custom pseudo-element like syntax when referring from outside the component. I think the fact that a component defined some pseudo element from a custom selector should be transparent to developers using a component

site/docs/references/custom-selectors.md
idoros: My bad

site/docs/references/custom-selectors.md
idoros: Once you group multiple selectors under a custom selector alias you get the same duplication of selectors anywhere you use it. This is not that bad.

But because that custom selector is exposed as a custom pseudo element, a user styling it from the outside has no idea that everytime he target that pseudo element he's selectors are multiplied.

This gets worse for every level you go up...

site/docs/references/custom-selectors.md
brucelawson: Yup, fair enough.

site/docs/references/custom-selectors.md
brucelawson: Yup, but my point is: does it matter? I believe that the performance of those selector is effectively free, and the resultant CSS, when minified and gzipped, is unlikely to bloat significantly. The cost of downloading a CSS file (compared with hero images, huge slabs of JS to parse then exectute) is pretty trivial (88K of the average page's 3.3 MB http://httparchive.org/interesting.php#bytesperpage). If it doesn't matter, don't mention it.

idoros: The problem is that the output is unpredictably big from a user position.

Example with 3 stylesheet each composition the other. Each describes a custom selector that expand the previous one with 2 groups. And at the end an application stylesheet that imports the top stylesheet and styling the pseudo-element that was generated by the custom selector:

```css
@namespace "A";
@custom-selector :--x .a1, .a2;
```
```css
@namespace "B";
:import {
    -st-from: "a.st.css";
    -st-default: A;
}
@custom-selector :--y .b1 A::x, .b2 A::x;
```
```css
@namespace "C";
:import {
    -st-from: "b.st.css";
    -st-default: B;
}
@custom-selector :--z .c1 B::y, .c2 B::y;
:--z { color: red; }
```
```css
@namespace "App";
:import {
    -st-from: "c.st.css";
    -st-default: C;
}
C::z { color: red; } /* only ruleset */
```

OUTPUT:
```css
.App--root .C--root .C--c1 .B--root .B--b1 .A--root .A--a1 { color: red; }
.App--root .C--root .C--c1 .B--root .B--b1 .A--root .A--a2 { color: red; } 
.App--root .C--root .C--c1 .B--root .B--b2 .A--root .A--a1 { color: red; }
.App--root .C--root .C--c1 .B--root .B--b2 .A--root .A--a2 { color: red; }
.App--root .C--root .C--c2 .B--root .B--b1 .A--root .A--a1 { color: red; }
.App--root .C--root .C--c2 .B--root .B--b1 .A--root .A--a2 { color: red; } 
.App--root .C--root .C--c2 .B--root .B--b2 .A--root .A--a1 { color: red; }
.App--root .C--root .C--c2 .B--root .B--b2 .A--root .A--a2 { color: red; }
```

brucelawson: Again, I'm not sure what the problem is. The resulting CSS is 8 lines; the Stylable source code is 10 lines. I think pointing out "footguns" should only be done in cases of extreme bloat/ performance problems. In this case, it's likely to deter people from using the library.

idoros: lines in stylable source- but only 1 actual CSS delaration

AnnieLR:
Reviewed and edited. Pushed my commit to this branch.

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,181 @@
...
+:--controlBtn { border: 1px solid grey; }
+/* 
+selector: .Comp__root .Comp__controls .Comp__btn:hover 
+*/
+:--controlBtn:hover { border-color: red; }
+```
+
+### Expose pseudo-element
+
+Custom selectors generate a [pseudo-element](./pseudo-elements.md). So, for example, [importing](./imports.md) a stylesheet into another stylesheet enables access to the `form-input` pseudo-element.
```
idoros: form-input should be controlBtn
and we should reference the prev example in the text

site/docs/references/custom-selectors.md
```diff
@@ -0,0 +1,181 @@
...
+
+The following example shows how a tree component exposes an icon.
+
+```css
+/* CSS */
+@namespace "Tree";
+@custom-selector :--icon .root > .icon;
+```
+
+Here you can use the icon `custom selector` from the outside just like you would use a CSS class selector or `pseudo-element`.
```
idoros: from the outside just like you would use any other `pseudo-element`.

site/docs/references/custom-selectors.md
idoros: @custom-selector :--media .image, .video;
/*
selector:
.Comp__root .Comp__image,
.Comp__root .Comp__video
*/
:--media {
...

site/docs/references/custom-selectors.md
AnnieLR: Done

site/docs/references/custom-selectors.md
AnnieLR: Done

site/docs/references/custom-selectors.md
AnnieLR: Done

PR #142, opened 2017-09-18
Qbllr/diag phase3

qballer:


src/stylable-processor.ts
```diff
@@ -368,6 +367,7 @@ export class StylableProcessor {
                     importObj.theme = parseTheme(decl.value);
                     break;
                 default:
+                    this.diagnostics.warn(decl,`"${decl.prop}" css attribute cannot be used inside :import block`, {word:decl.prop})
```
barak007: this applies only to none theme imports

src/stylable-transformer.ts
```diff
@@ -253,9 +248,12 @@ export class StylableTransformer {
         });
         rule.walkDecls(valueMapping.mixin, (node) => node.remove());
     }
-    replaceValueFunction(value: string, meta: StylableMeta) {
+    replaceValueFunction(node:any, value: string, meta: StylableMeta) {
```
barak007: update type to postcss.Node

PR #185, opened 2017-10-24
Annie/stylable intelligence

AnnieLR:
Wrote page
Added to TOC & Getting Started

### DIFF
--- a/site/docs/get-started.md
+++ b/site/docs/get-started.md
@@ -23,6 +23,8 @@ At build time, the preprocessor converts the Stylable CSS into flat, static, val
 
 3. While **Stylable** can work with any component framework, we use it with [React](https://facebook.github.io/react/docs/installation.html). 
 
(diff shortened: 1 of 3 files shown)


site/docs/get-started.md
```diff
@@ -23,6 +23,8 @@ At build time, the preprocessor converts the Stylable CSS into flat, static, val
 
 3. While **Stylable** can work with any component framework, we use it with [React](https://facebook.github.io/react/docs/installation.html). 
 
+4. Optionally you can install [**Stylable Intelligence**](./getting-started/stylable-intelligence.md), an extension providing IDE support for code completion and diagnostics.
```
arnonkehat: this needs to be a bit more verbose. we currently have an extension only for VSCode and will be making more in the future.

AnnieLR: This is the Getting Started topic for Stylable. The topic I added on Stylable Intelligence does include that. Repeat it here?

PR #209, opened 2017-11-28
pseudo classes with params

nadavwix:
Feature request: implement pseudo states with params.

Reason: Customization of data aware components ( AutoViews )

spec is found in [pseudo classes readme](https://github.com/wix/stylable/blob/nadav/pseudo-state-params/site/docs/references/pseudo-classes.md). 

- [x]  process implementation in stylable - tom
- [x]  implement in stylable - tom
- [ ]  process implementation in inteligence - ory
- [ ]  implement inteligence - ory
- [ ]  finalize readme - bruce & arnon
- [ ]  release notes - tom


tomrav:
new and improved syntax:

```css
/* 
* types and validators can be one and the same
* types/validarots are imported (optional)
* types/validators can have default values, e.g:  min vs min() useful for js validators
* values can have default values, e.g. -st-state(number, min(0), max(31)) 1;
* support multiple named params
* nesting validators - chain them one after the other, e.g. number(x()) === -st-state(number, x) 
*/


:import {
	-st-from: "stylable-validations";
	-st-named: multiplyOf, min, max, number;
	-st-default: -st-state;
}

:import {
	-st-from: "./my-validators.ts";
	-st-named: custom;
}

:-st-state(x, number, multiplyOf(3), min(0), max(30));

:-st-state(date) [data-date="$day-$month-$year"] {
	day: 		-st-state(number, min(0), max(31)) 1;
	month: 	custom(multiplyOf 3, min 0, max 30) 1;
	year: 	number(multiplyOf(3), min(0), max(30)) 1980;

	year: multiplyOf(3) 30;
}

.root {
	-st-states: date
}

.a {
	-st-states: date
}

.a:date(day 1, month 10, year 2001){

}
```

site/docs/references/pseudo-classes.md
```diff
@@ -38,12 +38,248 @@ To define custom pseudo-classes, or states, for a simple selector, you tell **St
...
+/* defining state */
+.token{
+    -st-states: fieldName(string);
+}
+
+/* customize fields with fieldName email */
+.token:fieldName(email){
+    color:lightBlue;
+}
+
```
tempit: Make string glob syntax identical to CSS.
fieldname($_id) does not look like it means 'ends with _id'
Make it fieldname($=_id) like CSS.

site/docs/references/pseudo-classes.md
```diff
@@ -38,12 +38,248 @@ To define custom pseudo-classes, or states, for a simple selector, you tell **St
...
+
+.cell:column(5x){
+    /* will let stylable know your gonna use the number current limits, with step value 5,
+    minValue of 0, maxValue of 10*/
+}
+
+```
+
+#### Boolean example
```
tempit: Make this first example, make clear it is the default. 
If it is possible to give an explicit Boolean parameter, give an example.

site/docs/references/pseudo-classes.md
```diff
@@ -38,12 +38,248 @@ To define custom pseudo-classes, or states, for a simple selector, you tell **St
...
+
+*all the future features described above are implemented using static processing.*
+
+*because of that you will have to declare them at build time in order to use them with dynamic css*
+
+```css
+.cell{
+    -st-states: column(number);
+    -st-states: done(number<0,100>);
+}
```
tempit: Two -st-states in one class. Is this correct syntax? Better if it were
 -st-states: column(number), done(number<0,100>);

tomrav: Just two examples side by side, no intend to support multiple directives (you can add multiple states for a single directive, and that should be good enough)

site/docs/references/pseudo-classes.md
tomrav: I agree that looks better, and more consistent with CSS.

site/docs/references/pseudo-classes.md
tempit: Yes, but in an example it looks like it's supposed to be correct syntax, which it isn't.
I'd either have two states in the same line, or (better?) two classes side-by-side.
You want this to be copy-paste-able, IMO.

brucelawson:
Some notes while trying to get my head around this stuff after xmas break/ sickness:

rename -st-state as -custom-state instead; 'cause any example that has an -st- prefix looks like it's part of Stylable core/ reserved word

brucelawson:
i don't know the functionality implied by this example - but "multiplyOf"  sounds alien to me. Is it "multiplyBy" so returns number*3, or "multipleOf" so returns true or false of number is perfectly divisible by 3?

src/pseudo-states.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    if (decl.parent && decl.parent.type !== 'root') {
+        const container = decl.parent;
+        if (container.type !== 'atrule') {
+            const sRule: SRule = container as SRule;
+            if (sRule.selectorAst.nodes && sRule.selectorAst.nodes.length === 1) {
+                const singleSelectorAst = sRule.selectorAst.nodes[0];
+                const selectorChunk = singleSelectorAst.nodes;
+
+                if (selectorChunk.length === 1 && selectorChunk[0].type === 'class') {
```
tomrav: refine with ido what about Elements

src/state-validators.ts
```diff
@@ -0,0 +1,237 @@
...
+
+            const res = value;
+            const errors: string[] = [];
+
+            if (validateValue && typeof value !== 'string') {
+                errors.push(validationErrors.string.STRING_TYPE_VALIDATION_FAILED(value));
+            }
+
+            if (validators.length > 0) {
+                validators.map(validatorMeta => {
```
tomrav: should be forEach

src/state-validators.ts
```diff
@@ -0,0 +1,237 @@
...
+            validators: StateArguments,
+            resolveParam: (s: string) => string,
+            validateDefinition,
+            validateValue) {
+
+            const res = value;
+            const errors: string[] = [];
+
+            const isNumber = !isNaN(value);
+            if (value !== 0 && !isNumber) {
```
tomrav: check value !== 0

PR #211, opened 2017-11-28
dynamic styles

nadavwix:
Feature request: add support for adding styles at runtime

reason: wix-code integration

- [ ] process tasks - barak&tom
- [ ] implement stylable support - tom
- [ ] implement stylable runtime support - tom
- [ ] process inteligence support - ory
- [ ] implement inteligence support - ory
- [ ] finalize readme - bruce & arnon
- [ ] release notes - tom


barak007:
The real issue here is how can we know the mapping of inner parts like `::item::deepItem`, they are not present in the current stylesheet runtime.

PR #212, opened 2017-11-29
DTS Generation

nadavwix:
Feature Request: Create DTS for Css files

Reason: better dev experience.

- [ ] finalize readme - arnon +bruce
- [ ] finalize DTS template - tom (use avi and amir)
- [ ] create DTS service - tom
- [ ] release notes - tom


barak007:
looks good.

I think that the optimal solution here is that typescript language services will generate the d.ts from the imported css via custom transformer, and we will be able to config it for IDEs.

otherwise we need to manage those files, and decide on:
* build location 
* source control

barak007:
This is a cleaner and more strict way to express the stylsheet api.


```ts



/* this will receive the component props and we dont know the type here. we only need  data-* */ 

export interface PartialProps {
    "className"?: string
    [key: string]: any;
};

export interface StylableProps {
    "className"?: string
    "data-ns-selected"?: boolean
};

export type StateMap = { [key: string]: never | null } & {
    root: {
        selected?: number
        hover?: number
    } | null
}

export type StyleSheet = {
    <ClassName extends keyof StyleSheet>(className: ClassName, states?: StateMap[ClassName], props?: PartialProps): StylableProps
    root: 'a--root'
    b: 'b--b'
    c: 'c--c'
}


var s = {} as StyleSheet;


```

PR #215, opened 2017-11-30
changed home page - still no GIFs

arnonkehat:


brucelawson:
It was a good call to change the homepage.  I was talking F2F with a developer friend last night, who didn't "get" Stylable. When I explained it with the new copy, he understood it and was excited!

PR #234, opened 2018-01-04
Annie/formatter docs

AnnieLR:
Added to cheatsheet - just as an item no syntax.

### DIFF
--- a/site/docs/references/formatters.md
+++ b/site/docs/references/formatters.md
@@ -9,7 +9,7 @@ Formatters are functions that return a single CSS declaration value. They can re
 >**Note**  
 >Currently, any argument passed through to a formatter is of type `string`. We are in the process of adding support for more complex types.
 
-For example a `calc-font-size` formatter can return a different value for the font size depending on the modifier/argument applied.
+For example a `calc-font-size` formatter can return a different value for the font size depending on the provided argument.
 
 
 ```js
--- a/site/docs/references/formatters.md
+++ b/site/docs/references/formatters.md
@@ -6,7 +6,7 @@ layout: docs
 
 Formatters are functions that return a single CSS declaration value. They can receive arguments, process them and return the value.
 
-For example a `font-size` formatter can return a different value for the font size depending on the location.
+For example a `font-size` formatter can return a different value for the font size depending on the tag applied.
 
 
 ```js
--- a/site/docs/references/formatters.md
+++ b/site/docs/references/formatters.md
@@ -4,129 +4,134 @@ title: Formatters
 layout: docs
 ---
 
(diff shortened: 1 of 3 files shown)


site/docs/references/formatters.md
```diff
@@ -6,7 +6,7 @@ layout: docs
 
 Formatters are functions that return a single CSS declaration value. They can receive arguments, process them and return the value.
 
-For example a `font-size` formatter can return a different value for the font size depending on the location.
+For example a `font-size` formatter can return a different value for the font size depending on the tag applied.
```
tomrav: This should be modifier/argument instead of tag

site/docs/references/formatters.md
```diff
@@ -55,7 +55,9 @@ For example a `font-size` formatter can return a different value for the font si
 
 ## Formatters with variables
 
-You can use formatters with variables:
+You can use formatters with variables.
```
tomrav: This feels pretty detached - might want to buff it up a bit

site/docs/references/formatters.md
```diff
@@ -88,7 +90,9 @@ You can use formatters with variables:
 
 ## Nested formatters
 
-You can use nested formatters:
+You can use nested formatters.
```
tomrav: Same comment as above, feels detached.

site/docs/references/formatters.md
```diff
@@ -126,9 +130,3 @@ You can use nested formatters:
         return Math.PI * multiplyBy;
```
tomrav: replace this with the following example, and replace usage in the rest of the example:

```js
/* ./sin.js */
module.exports = function(degree) {
    return Math.sin(Number(degree));
}

File: site/docs/references/formatters.md [hunk 5]
```diff
@@ -55,7 +55,9 @@ For example a `font-size` formatter can return a different value for the font si
 
 ## Formatters with variables
 
-You can use formatters with variables:
+You can use formatters with variables. 
+
+In this example the CSS imports the same formatter as the previous example, `cal-font-size`, but the variable `baseFontSize` is added to the calculation.  
 
 ```css
     :import {
```
tomrav: in the var definition - remove the px, as it will be added from the formatter (otherwise we'll get something like `12pxpx`)

site/docs/references/formatters.md
```diff
@@ -6,7 +6,7 @@ layout: docs
 
 Formatters are functions that return a single CSS declaration value. They can receive arguments, process them and return the value.
```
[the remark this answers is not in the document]
AnnieLR: Don't think we should add link to a pull request. It will probably be broken when pull request is merged.

Issue #71, opened 2017-08-16
support multiple selectors in createClassSubsetRoot


barak007:
source:
```css
.a { 1 }
.a:hover { 2 }
.a, .b, .a .c { 3 }

.x {
    -st-mixin: a;
}
```

target:
```css
.a { 1 }
.a:hover { 2 }
.a, .b, .a .c { 3 }

.x { 1 } 
.x:hover { 2 } 
.x, .x .c { 3 } /* .b is removed since its not part of the mixin '.a' */
```

`.x, .x .c { 3 }` - not outputted today - bug

tomrav:
Resolved in v5.1.0 
https://github.com/wix/stylable/releases/tag/v5.1.0

Issue #229, opened 2017-12-27
TS Mixin renders an extra whitespace before pseudo elements


tomerlichtash:
**TS Mixin**
```
module.exports = (content: any) => {
  return {
    "::before": {
      content: `\"___${content}___\"`
    }
  };
}
```

**CSS**
```
Item::item-title{
  -st-mixin: slidingDashMixin();
}
```

**Result: Extra whitespace before the pseudo element:**
```
.themea3567016190--root .item623790088--root .item623790088--item-title ::before {
  content: "yyyyyyyy___-___yyyyyyyyyyyyy";
}
```

tomrav:
This is behaving as expected.
By writing the inner selector starting immediately with pseudo selector (`::before`) you get a descendant selector.

If you wish to concatenate your classes (`.item623790088--item-title::before`) add an `&` to your selector: `"&::before": {`

Issue #228, opened 2017-12-27
Multiple mixins overridden by first declared mixin when using TypeScript


tomerlichtash:
Using Typescript Mixins will prefer the first declaration of the Mixin over the others. This is when using TypeScript mixins only. JS mixins behave normally.

Take this case, for example:
```
List::list-item:first-child {
  -st-mixin: RoundBadge(100, value(st_real_white), value(color1), 'hmm', value(badgeTextSize), -20px, 0, 0, -20px);
}
List::list-item {
  -st-mixin: RoundBadge(70, value(st_real_white), value(color2), 'hmpf', value(badgeTextSize), -20px, 0, 0, -20px);
}
List::list-item:last-child {
  -st-mixin: RoundBadge(30, value(st_real_white), value(color3), 'ha', value(badgeTextSize), 0, 500px, 0, 500px);
}
```
In the above case the RoundBadge mixin used for :first-child with its params will be used in all other RoundBadge declarations.

This is not unique for similar namespaces; the following will enforce the first mixin on the other, even though they are in different scopes:
```
Item::description {
  -st-mixin: RoundBadge(100, value(st_real_white), value(color1), 'hmm', value(badgeTextSize), -20px, 0, 0, -20px);
}
List::list-item {
  -st-mixin: RoundBadge(70, value(st_real_white), value(color2), 'hmpf', value(badgeTextSize), -20px, 0, 0, -20px);
}
```

tomerlichtash:
Closing issue - turns this originates in the way the mixin was constructed (the reason is some test to get named params):

```
import {stString,stSize, stNumber, stColor, CSSObject, stCssFrag} from 'stylable';

type BadgePropTypes = [
  stString,
  stColor,
  stColor,
  stNumber,
  stNumber,
  stNumber,
  stNumber,
  stNumber,
  stNumber
];

module.exports = ({
  0: proportions,
  1: fg,
  2: bg,
  3: content,
  4: size,
  5: top,
  6: right,
  7: bottom,
  8: left
}: BadgePropTypes) => {
  module.exports = (params: BadgePropTypes): any => {
    const prps: any = `${proportions}px`;

    return {
      "font-size": size,
      "&::before": {
        "line-height": prps,
        "border-radius": prps,
        content: `"${content}"`,
        width: prps,
        height: prps,
        background: bg,
        color: fg,
        top, right, left, bottom
      }
    };
  }
}
```

Issue #226, opened 2017-12-25
CSS Mixin within a CSS Mixin will not transpile correctly


tomerlichtash:
```project.st.css``` imports ```style1``` and declares ```Item::title {-st-mixin: Style1}```.
```style1.st.css``` imports ```no-select``` and declares ```Style1 {-st-mixin: no-select}```.

**Expected result**: Style1 will contain rules from the original style definition, in which ```no-select``` will be embedded.

**Actual result**: The latter does not happen.

tomrav:
Resolved in v5.1.0 
https://github.com/wix/stylable/releases/tag/v5.1.0

brucelawson:
"CSS Mixin within a CSS Mixin" < from now on, this concept is to be referred to as a "Mix-inception".  Let it be so.

PR #237, opened 2018-01-07
Mixin with params

### LINKED ISSUE #231: functions with named parameters
Currently Stylable supports JS mixins with argument list and this is fine for small functions with no optional params, but when using functions or CSS mixins with several params and no specific order it would be nice to pass just the mandatory/overridden params.

I couldn't find any CSS syntax / future spec that offer such input (@brucelawson asked in the [Houdini project](https://github.com/w3c/css-houdini-drafts/issues/551) and currently there is no answer there).

I think it should be something like this:

```css
.root {
    -st-mixin:
        modifyFunc(
            color1: green,
            color2: red
        );
}
``` 
 
This works nicely with `postcss` and `postcss-value-parser`, and I couldn't find any objection to the syntax in the CSS spec, however, any IDE that I tried it in raised an error on the colon between the param name and value.

We could support named params with the first space as separator between name and value:

```css
.root {
    -st-mixin:
        modifyFunc(
            color1 green,
            color2 red
        );
}
``` 

@nadavov, @barak007  - What do you think?

barak007:
this resolves #231


idoros:
should fix:
#71 
#121 
#226 
#228 
#229

PR #239, opened 2018-01-09
refactored the doc for mixins to include both CSS and JS/TS mixins

tomrav:


site/docs/references/mixin-syntax.md
```diff
@@ -1,98 +1,160 @@
----
-id: references/mixin-syntax
-title: Mixins
-layout: docs
----
```
arnonkehat: we still need this part, right?

tomrav: Yeah, please add

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+
+Here are some examples of when you can use mixins:
+
 * Presets/Variants - create reusable pieces of styling CSS
 * Layouts - easily describe complex layouts
 * Effects - easily describe complex effects
-* Macros - use JS to define the CSS macros you need
+* Macros - use code to define the CSS macros you need
+
+Mixins can be applied using either CSS or JavaScript, and can recive parameters.
```
idoros: I think this should be mixed-in with the first sentence:

**Stylable** mixins enable you to reuse complex styles and CSS behaviors, defined in CSS or Javascript, and apply them to any ruleset during build time.

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+/* CSS output */
+...
+.someClass {
+    color: green; /* from local mixin */
+    background: yellow; /* from local mixin */
+}  
 ```
 
-## Defining mixins in CSS
+Here is an example of a **Stylable** CSS file that is imported and mixed into the  classes of a different stylesheet. The `.rootMixedIn` class as a stylesheet and `classMixedIn` as a class.
```
idoros: double space... is there any lint for text / markdown?

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+```
 
-/* file a.st.css */
-.a{
-    color:red;
+``` css
+/* CSS file - example.st.css - imports the above mixin */
+:import {
+    -st-from: './mixins.st.css';
+    -st-default: MyComp;
```
idoros: I would change `MyComp` to `MixRoot`

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+    color: green; /* from class mixin */
+}
+```
 
-## Applying multiple mixins
+### CSS mixin with parameters and variables
 
-You can apply multiple mixins in the same line.
+CSS mixins can accept named parameters in the following format:
+ `mixin(variableName valueOverride)`.
```
idoros: add a second parameter to the syntax example:

`mixin(variableName valueOverride, variableName2 valueOverride2)`

It might be also nice to show here that it works in multiline aswell:

```css
mixin(
    variableName valueOverride, 
    variableName2 valueOverride2
)
````

File: site/docs/references/mixin-syntax.md [hunk 6]
```diff
@@ -1,124 +1,253 @@
...
+.targetClass {
+    background: orange; /* from mixin with override */
 }
 ```
 
-## How mixins are applied
+## JavaScript mixins
 
-Mixins can add CSS declarations to the CSS ruleset to which they are applied:
+JavaScript mixins allow you to create complex structures in CSS based on the arguments passed to the mixin and the mixin logic.
```
idoros: the "and the mixin logic" at the end is redundant

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
-## How mixins are applied
+## JavaScript mixins
 
-Mixins can add CSS declarations to the CSS ruleset to which they are applied:
+JavaScript mixins allow you to create complex structures in CSS based on the arguments passed to the mixin and the mixin logic. 
 
-* Rules are added at the position in the CSS where the `-st-mixin` is declared.
-* Any selectors that are appended as a result of the mixin are added directly after the ruleset that the mixin was applied to.
-* Multiple mixins are applied according to the order that they are specified.
+A JS mixin returns a CSS fragment which can be a single declaration and its value, multiple declarations, multiple rulesets, or any combination of these.
```
idoros: "single declaration and its value" - declaration is a `prop:value` definition, "and its value" is not needed.
I would shorten this into:

A JS mixin returns a CSS fragment which can contain multiple declarations with optional sub style-rules

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+```
+
+## How mixins are applied
+
+Mixins can add CSS declarations to the CSS ruleset to which they are applied.
+
+Rules are added at the position in the CSS where the `-st-mixin` is declared.
+Any selectors that are appended as a result of the mixin are added directly after the ruleset that the mixin was applied to.
+
+You can apply multiple mixins in either CSS or JavaScript, or both seperated by comma `-st-mixin: mixinA, mixinB`.
```
idoros: "in" or should it be "from"?

site/docs/references/mixin-syntax.md
```diff
@@ -1,124 +1,253 @@
...
+
+### Mixin usage with special characters
+You can escape special characters by wrapping them with quotes or using a backslash (`\`). 
+
+Example:
+```css
+.x { /* use quotations to include comma */
+    -st-mixin: mix(300, "xx,x"); /* ["300", "xx,x"] */
+}
+.y { /* escape slashes */
```
idoros: this are not escape slashes. we use slash to escape special characters. here the escape is used to have quotation chars as part of the parameter

PR #251, opened 2018-01-21
fix #250 - Custom pseudo classes after native pseudo classes

### LINKED ISSUE #250: Custom pseudo classes after native pseudo classes do not transpile
This example will transpile correctly:
```
.typedComp::innerPart {
  color: red;
}
```

However, adding a pseudo class will cause end with untranspiled selector:

```
.typedComp:hover::innerPart {
  color: red;
}
```

will end up as ```...:hover::primary``` in the transpiled css.

CC: @barak007

barak007:


barak007:
Wrong issue tagged. Closing and will re-open with the correct fix.

PR #284, opened 2018-02-26
Stylable proxy - for testing

shlomitc:
This PR adds a very basic proxy mechanism for CSS classes, in order to apply it many test frameworks.

Usage example in `jest` configuration:
```
"moduleNameMapper": {
  "\\.(st.css)$": "stylable/stylable-proxy"
}
```


tests/utils/stylable-proxy.ts
```diff
@@ -0,0 +1,21 @@
+import {create} from '../../src/runtime';
```
barak007: Please create a function compatible to the create function in runtime.ts and export it in order to use it in the runtime test suite.

tests/stylable-proxy.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  before(() => {
+    require.extensions['.css'] = function mockCss(module) {
+      module.exports = stylableProxy;
+    };
+  });
+
+  it('should apply the proxy function and preserve stylable', () => {
+    const css = require('./test-fixtures/simple.st.css');
+    expect(css.a).to.equal('a');
+    expect(css('root', {state: true}, {prop: 1})).to.eql({
```
shlomitc: @barak007 
Now I see that I didn't test the `default` behavior:

```
expect(css.default('root', {state: true}, {prop: 1})).to.eql({
  className: 'root',
  "data-namespace-state": true
});
```

It fails - how should it be handled in my proxy?

tests/utils/stylable-proxy.ts
shlomitc: you mean just wrapping what I export as a function?

shlomitc: you mean just wrapping what I export as a function?

tests/stylable-proxy.spec.ts
```diff
@@ -0,0 +1,25 @@
+import 'mocha';
+import {expect} from 'chai';
+import stylableProxy from './utils/stylable-proxy';
+
+describe('stylable-proxy', () => {
```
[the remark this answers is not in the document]
shlomitc: Running my proxy against all the tests in `runtime.spec.ts` means I need to implement the entire logic. 

for example, this test will fail `should map classNames to locals with fallback to global classes`

```
AssertionError: expected { className: 'root local global' } to deeply equal { Object (className) }
+ expected - actual

 {
-  "className": "root local global"
+  "className": "namespace--root namespace--local global"
 }
```

So I don't think it's a good idea to treat them with the same tests

PR #285, opened 2018-02-26
rule shaking

idoros:
Improve rule shaking support for 3rd party stylable applications


idoros:
fix #285 when importing used components through an "unused" library index file

barak007:
* changed behavior of resolvePath so the context is defaults to the project root

PR #335, opened 2018-05-29
* Created test kit project for webpack builds

barak007:


idoros:
@stylable/webpack-test-kit?

PR #344, opened 2018-06-11
Stylable:

barak007:
* Fixed resolution of 3rd party packages that have different context
StylableWebpackPlugin:
* Added test for resolution of nested 3rd party packages


wilomgfx:
Would this pr fix my problem ?
https://github.com/wix/stylable/issues/345

PR #378, opened 2018-07-10
Core: remove deprecated features

tomrav:
> Do **NOT** merge to master, contains breaking changes.

---------------

Removes deprecated features:
- `-st-theme`
- `-st-variant`
- `-st-compose`
- core bundler `bundle.ts`


packages/stylable-webpack-plugin/test/e2e/dev-mode-warnings-project.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+            const computedStyle = getComputedStyle(document.body, '::before');
+            return {
+                color: computedStyle.backgroundColor,
+                content: computedStyle.content
+            };
+        });
+
+        expect(values.color).to.eql('rgb(255, 0, 0)');
+        // tslint:disable-next-line:max-line-length
+        expect(values.content).to.eql(`"Invalid CSS class assignment of '.o0--root', target node is missing extended class '.o1--root'"`);
```
idoros: can we change the warning to show the original class names? and maybe change it to something in the lines of:

packages/stylable/src/selector-utils.ts
```diff
@@ -260,3 +260,41 @@ export function isChildOfAtRule(rule: postcss.Rule, atRuleName: string) {
...
+                prop: 'content',
+                value: message
+            }),
+            postcss.decl({
+                prop: 'width',
+                value: `200px !important`
+            }),
+            postcss.decl({
+                prop: 'height',
+                value: `200px !important`
```
idoros: 200x200?

packages/stylable/src/stylable-meta.ts
```diff
@@ -53,18 +53,14 @@ export interface Imported {
...
     rule: postcss.Rule;
     fromRelative: string;
 }
 
 export interface StylableDirectives {
     '-st-root'?: boolean;
-    '-st-compose'?: Array<ImportSymbol | ClassSymbol>;
     '-st-states'?: any;
     '-st-extends'?: ImportSymbol | ClassSymbol | ElementSymbol;
-    '-st-theme'?: boolean;
```
idoros: need to check that intelligence won't break...

packages/stylable/src/stylable.ts
```diff
@@ -66,6 +67,7 @@ export class Stylable {
         protected scopeRoot: boolean = true,
         protected resolveOptions: any = {},
         protected optimizer?: StylableOptimizer,
+        protected mode: 'production' | 'development' = 'production',
```
idoros: doesn't the plugin defaults to development? is this set to production in order to avoid the warning rule in tests?

packages/stylable/src/stylable-meta.ts
tomrav: Either way it'll be a major change and we'll have to go through the intelligence. Don't think this should be a big issue.

packages/stylable/src/stylable.ts
tomrav: I don't think that having visual warnings pop up for a user without him asking for `dev` mode explicitly would be a good idea, hence defaulting to `production`.

packages/stylable-webpack-plugin/test/e2e/dev-mode-warnings-project.spec.ts
tomrav: Sure, sounds good.

packages/stylable/src/selector-utils.ts
idoros: I need to look at how it shows on the screen, but I had a small floating message in mind, not a 200x200 block. Ideally, like @barak007 said, this should be coupled with dev run-time code that will move this message to the console.

packages/stylable/src/stylable.ts
idoros: I agree, but the default for the webpack plugin is `dev`, so if we agree that `production` is a better default, then lets change the webpack-plugin as well (just ask @barak007).

tomrav: I don't believe the `webpack-plugin` itself has a default. We have a default in `stylable-scripts` for the `start`/`build` scripts which get resolved to `development`/`production` modes respectively.

If someone were to just bring `stylable` and `stylable-webpack-plugin` as dependencies, whatever they'll write in their webpack config will be the default.

barak007:
Closed in favor of a new PR from the correct branch `scoped-stylable`.

Issue #217, opened 2017-12-01
Simplify root scoping


idoros:
# Root scoping

> Notice: this proposition is still WIP and not ready yet:  
> * consider themes
> * anything else?

Currently the way Stylable handles scoping is not ideal. The purpose of this thread is to explain the issues related to the current behavior and suggest alternatives.

## Current state

Every selector in Stylable is transformed to resolve global names, remove style API abstractions (custom pseudo elements and states) and to prevent "leaking" of not locally namespaced parts outside the scope of the component (like [tag-selectors](https://stylable.io/docs/references/tag-selectors) and imported [shared classes](https://stylable.io/docs/guides/shared-classes) and [component-variants](https://stylable.io/docs/guides/component-variants)).

```css
@namespace "a";

/*
The root class is added to prevent the stylesheet from affecting div elements outside of the component (this has an obvious problem of leaking styles to nested scopes and it is recommended to "anchor" the selector: `.root > div`)

@selector: .a__root div
*/
div {}

/*
root is added to keep the the specificity of normal classes above tag elements

@selector: .a__root .a__local
*/
.local {}

:import {
    -st-from: ".b.st.css";
    -st-default: B;
    -st-named: innerB;
}
/*
Like tag selectors, imported shared classes and component variants needs to be scoped to the root in order to prevent "leaking"

@selector: .a__root .b__innerB
*/
.innerB {}
```

## Why is this good? bad? and alternatives

* **good**: 
    * standalone selectors like `tags` and `shared classes` are kinda scoped to the component automatically
* **bad**: adding scope to none local symbols is a magical solution that adds complexity
    * causes an **unintuitive** result in target CSS   
    * **adds specificity** to the result selector that in turn complex the entire system
    * (AND) it only solve half of the scoping issue - everything nested in the component is still affected
    * causes CSS/portal issues right out of the box (because **every** selector is a nested selector)
* **alternative**: stop root scoping!  
since almost every local is namespaced and doesn't need the root class for scoping. It might be better off to stop adding the root class. 
    * specificity is simpler to understand
    * non local symbols are basically global, but it can be solved by a simple rule that warn/error when such symbols are not "anchored" to a local class: `.root > div`

nadavwix:
this requires build linting, 

we cannot allow components with a tag selector
we cannot allow components that modify a global class


this means we need to differentiate themes from component styles:

```css
:import{
   -st-from:'./my-comp.st.css'
   -st-default: MyComp;
   -st-named: CompVariant;
[...]

PR #387, opened 2018-07-16
Core: remove root scoping

tomrav:
Per issue #217


barak007:
This is also removes old optimize in transformer. Looks good from here.

PR #438, opened 2018-10-12
react-scripts: upgrade stylable's version from upstream

AviVahl:
- requires newer beta of html-webpack-plugin
- removed webpack4 forks of plugins and now using react-dev-utils again
- removed dead test script
- updated boiler app to include typescript version of service worker installation. cleaned component types.


AviVahl:
should probably be noted that their default chunk configuration broke stylable's runtime style renderer, so I disabled it.
webpack config had:
``` 
optimization: {
    // Automatically split vendor and commons
    // https://twitter.com/wSokra/status/969633336732905474
    // https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366
    splitChunks: {
      chunks: 'all',
      name: false,
    },
    // Keep the runtime chunk seperated to enable long term caching
    // https://twitter.com/wSokra/status/969679223278505985
    runtimeChunk: true,
  }
```

PR #446, opened 2018-11-06
Add 'var' to the list of native css functions

tomrav:


wtfil:
Just wanted to create the same issue. Can you pls publish new version?

tomrav:
> Just wanted to create the same issue. Can you pls publish new version?

I will be releasing a new version today. :)

Issue #450, opened 2018-11-15
`unscoped native element` warning


hepiyellow:
I have this use case where I want to apply theme namespace:
 
> `theme.st.css`
```css
@namespace "BackofficeTheme";
```

> `someComp.st.css`
 ```
 :import {
   -st-from: "../theme.st.css";
   -st-default: BOTheme;
 }
 ```
 ![image](https://user-images.githubusercontent.com/18248989/44086713-17d87b26-9fc6-11e8-80e7-f674cce9854f.png)
 
 I am Getting
 `[stylable] unscoped native element "BOTheme" will affect all elements of the same type in the document`. 

Why is this?
It is not a native element.

hepiyellow:
1. Maybe it should just be fixed to `unscoped custom element`.
1. By the way, what if I DO want to affect all elements of the same type in the document. Can I disable the warning?

tomrav:
1. I agree the error should be more exact if you've used native/custom element, I'll be sure to fix it before our next release.
2. Short term, you should use the `@st-scope` feature once it's available (Soon. ™️). Later on, we'd like to expand our diagnostics in general to allow more user flexibility.

tomrav:
There's currently no way to turn off diagnostics in Stylable (we're hoping to get to it at some point).

Regardless of that, how are these warnings from build getting to the browser? Are you running webpack-dev-server?

As an aside, I hope to release the new `@st-scope` feature today, so you could wait and use that instead.

tomrav:
Resolved in #455 
The error message has been made more clear, and `@st-scope` feature introduced to allow for controlled scoping.

Issue #445, opened 2018-11-05
Silence "unscoped native element" warning


JoMarton:
Getting a LOT of warnings in `wix-ui-core` when running `npm run start`:

```
WARNING in Stylable: /Users/marton/code/wix-ui/packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css:350:1: unscoped native element "BackofficeTheme" will affect all elements of the same type in the document

  348 | }
  349 |
> 350 | BackofficeTheme .underlined.dark:active:not(:disabled) {
      | ^
  351 |   background-color: transparent;
  352 |   border-color: transparent;
```

I get one of these blocks for every class in every stylesheet. This scoping is by design currently so having the console fill up like that is very annoying.

I suggest:
- flag to silence these warnings
- I love how descriptive this is but maybe formatting in a more concise way?

tomrav:
This is related to a feature we're intending on adding to Stylable soon, #411.

As a separate issue, we should allow some sort of configuration to disable specific warnings.

tomrav:
Resolved in #455 
The error message has been made more clear, and `@st-scope` feature introduced to allow for controlled scoping.

Issue #411, opened 2018-08-14
Theme namespace


lbelinsk:
I have this use case where I want to apply theme namespace:

```
:import {
  -st-from: "../theme.st.css";
  -st-default: BOTheme;
}
```

![image](https://user-images.githubusercontent.com/18248989/44086713-17d87b26-9fc6-11e8-80e7-f674cce9854f.png)

I am Getting 
`[stylable] unscoped native element "BOTheme" will affect all elements of the same type in the document`. Other than that warning, it will be great to have some nice syntax to apply only once at the top of the stylesheet that I am "root scoped" to this theme, instead of repeating this over and over for each rule in the stylesheet.

lbelinsk:
Another issue that might get resolved automatically by fixing this:

// button.st.css
```
BOTheme .premium {
  background-color: #AA4DC8;
  border-color: #AA4DC8;
  color: white;
}

```

// dropdown.st.css
```
:import {
  -st-from: "../button/button.st.css";
  -st-named: premium;
}

BOTheme Dropdown::actionButton {
  -st-mixin: premium; // Not working since premium class is namespaced under the theme in button.st.css
}
```

tomrav:
I'm starting work on this feature, we've come up with the syntax below.
It is limited to a single level (no inner nesting), but will allow for multiple different scopes in the same file. By using `@st-scope` the warning you've been getting will be ignored and mixins should work.

Appreciate any comments you might have.

```css
:import {
    -st-from: "some-package/BOTheme.st.css";
    -st-default: BOTheme;
}

@st-scope BOTheme {
    .root{}
    .btn {}
};
```

mykas:
Looks pretty good, Tom.  @hepiyellow  what do you think?

lbelinsk:
@tomrav just to make sure I understand, all of the rules inside 
```
@st-scope BOTheme { ... }
```
Will be scoped, just like we do today:
```
BOTheme  .someClass {}
BOTheme  .someClass2 {}
```

If so, sounds great

And what do you mean by `no inner nesting`? Can you give an example of an unsupported syntax?

lbelinsk:
Also, what about supporting this:
```
BackofficeTheme .button,
BackofficeTheme.button {
    .... rules....
}
```

Placing the theme on the component itself, not on some wrapper

mykas:
About inner nesting. Maybe he means that no such cases:
```css
@st-scope BOTheme { 
    @st-scope BOTheme { ... }
 }

[...]

PR #455, opened 2018-11-20
[WIP] @st-scope

tomrav:
Add a new syntax feature to Stylable - `@st-scope`.

As referenced in issues #411 #445 #450.

Further information will be included in the readme to come.

### DIFF
--- a/packages/core/src/selector-utils.ts
+++ b/packages/core/src/selector-utils.ts
@@ -259,7 +259,7 @@ export function fixChunkOrdering(selectorNode: SelectorAstNode, prefixType: Sele
     });
 }
 
-export function isChildOfAtRule(rule: postcss.Rule, atRuleName: string) {
+export function isChildOfAtRule(rule: postcss.Container, atRuleName: string) {
     return rule.parent && rule.parent.type === 'atrule' && rule.parent.name === atRuleName;
 }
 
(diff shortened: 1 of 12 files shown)


packages/core/tests/scope-directive.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+                @st-scope root {
+                    .part {}
+                    .otherPart {}
+                }
+            `,
+                { from: 'path/to/style.css' }
+            );
+
+            const rules = meta.ast.nodes!;
+            expect((rules[1] as Rule).selector).to.equal('.root .part');
```
idoros: why does it transform the AST at the process phase? shouldn't the change only be in the transtorm?

tomrav: By removing the syntax sugar during processing we're ending up with selectors that are still Stylable symbols (untransformed), this makes the processing result already partly usable.

Similar to how we behave with imports/vars/custom selectors, we remove our own custom syntax from the ast where possible, leaving the final transformation up to the transformer.

packages/core/src/selector-utils.ts
```diff
@@ -108,6 +108,19 @@ export function isRootValid(ast: SelectorAstNode, rootName: string) {
 
 export const createSimpleSelectorChecker = createChecker(['selectors', 'selector', ['element', 'class']]);
 
+export function isSimpleSelector(selectorAst: SelectorAstNode) {
+    const checker = createSimpleSelectorChecker();
+    let isSimple = true;
+
+    traverseNode(selectorAst, node => {
```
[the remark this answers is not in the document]
tomrav: Done 👍

PR #476, opened 2018-12-18
[Bug] Failing state resolution in complex selector

tomrav:
### Source example
```css
/* imported.st.css */
.root { -st-states: disabled; }
```

```css
/* entry.st.css */
:import {
    -st-from: "./imported.st.css";
    -st-default: Base;
}

.class { -st-extends: Base; }
.modifier { -st-extends: Base; }

.class:disabled.modifier {} /* works */
.modifier.class:disabled {} /* doesn't work */
```

### Expected
```css
.entry--class[data-imported-disabled].imported--modifier {}
.imported--modifier.entry--class[data-imported-disabled] {}
```

### Actual
```css
.entry--class[data-imported-disabled].imported--modifier {} /* works */
.entry--modifier.imported--class:disabled {} /* broken */
```

### DIFF
--- a/packages/core/tests/pseudo-states.spec.ts
+++ b/packages/core/tests/pseudo-states.spec.ts
@@ -1739,6 +1739,52 @@ describe('pseudo-states', () => {
 
             });
         });
+
+        describe('concatenated classes in a complex selector', () => {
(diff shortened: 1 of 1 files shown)


tomrav:
Resolved in #839

PR #482, opened 2019-01-09
Add support for CSS custom properties (CSS Vars)

### LINKED ISSUE #247: CSS vars support
Because CSS vars are not namespaced by design, it would be nice to have some way to distinguish between conflicting and shared vars. Stylable can offer support for auto namespaced CSS vars, but it means the code needs to import vars in order to use them:

```css
:import {
    -st-from: "./panel-with-dropdown.st.css";
    -st-default: Panel;
}
:import {
    -st-from: "./dropdown.st.css";
    -st-named: --optionColor;
}
.panel {
    -st-extends: Panel;
    --optionColor: blue; // namespaced according to dropdown scope
}
```

tomrav:
Resolves #247, see [comment](https://github.com/wix/stylable/issues/247#issuecomment-449982509) for syntax details.


tomrav:
This will not be merged to master just yet, but to a new branch for a new major version.

PR #507, opened 2019-02-24
fix edge function cache

barak007:


packages/runtime/src/css-runtime-stylesheet.ts
```diff
@@ -77,7 +77,11 @@ export function create(
 
   Object.setPrototypeOf(stylable_runtime_stylesheet, locals);
 
+  // EDGE CACHE BUG FIX
```
AviVahl: link to issue in comment https://github.com/Microsoft/ChakraCore/issues/5970

idoros:
```suggestion
  // EDGE CACHE BUG FIX: https://github.com/Microsoft/ChakraCore/issues/5970
```

PR #707, opened 2019-07-17
[core, schema-extract] Add cssDocs support

tomrav:
Expose a utility method in `@stylable/core` called `getCssDocsForSymbol`, that accepts a `StylableSymbol` (`class` or `element` for the moment), and attempts to extract the docBlock written above the symbol node.

These cssDocs are automatically added to the output from `@stylable/schema-extract`.

### DIFF
--- a/packages/core/src/stylable-meta.ts
+++ b/packages/core/src/stylable-meta.ts
@@ -22,7 +22,7 @@ export class StylableMeta {
     public parent?: StylableMeta;
     public transformDiagnostics: Diagnostics | null;
     public scopes: postcss.AtRule[];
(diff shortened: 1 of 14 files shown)


packages/core/src/stylable-meta.ts
```diff
@@ -22,6 +22,7 @@ export class StylableMeta {
     public parent?: StylableMeta;
     public transformDiagnostics: Diagnostics | null;
     public scopes: postcss.AtRule[];
+    public mappedSimpleSelectors: Record<string, MappedComment>;
```
tomrav: change `MappedComment` name

PR #723, opened 2019-07-28
Fix cli compat output

barak007:
* Add default export to cli output in compat mode
* Add basic build manifest generation

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -2,6 +2,7 @@
 
 import { Stylable } from '@stylable/core';
 import fs from 'fs';
+import path from 'path';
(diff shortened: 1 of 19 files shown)


packages/cli/src/cli.ts
```diff
@@ -153,7 +165,8 @@ build({
     outputCSSNameTemplate: cssFilename,
     optimize,
     compat,
-    minify
+    minify,
+    manifest: manifest ? path.join(rootDir, outDir, manifestFilepath) : undefined
```
tomrav: According to this line, should `manifestFilepath` be `manifestFilename`?

tomrav:
Also missing the README update.

PR #724, opened 2019-07-31
fix(core): issue with trailing comma in mixin invocation

AndriiTsarenko:


tomrav:
Thanks for the PR, I'll take a look at it early next week. :)

tomrav:
I've opened PR #755 which solves the same issue, thanks for suggesting the fix.

As for the change in the `package.json` file, you don't actually need it, and can run `yarn test -w` to enter watch mode instead.

PR #748, opened 2019-08-14
Added processDeclarationFonts

DeDuckProject:
Stylable process now collects used fonts (same as it collects used URLs).
The fonts are then exposed on meta.fonts.


tomrav:
Will pick up this work at a later date

PR #757, opened 2019-08-29
Add flag to CLI accepting require hooks to run

tomrav:
This will allow to pass the CLI `@ts-tools` require hook handler to run typescript (mixins, custom values, etc) from source.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -147,9 +147,9 @@ const {
 
 log('[Arguments]', argv);
 
-for (const r of requires) {
-    if (r) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/cli.ts
```diff
@@ -134,11 +141,18 @@ const {
...
     minify,
     manifestFilepath,
-    manifest
+    manifest,
+    require: requires
 } = argv;
 
 log('[Arguments]', argv);
 
+for (const r of requires) {
```
barak007: change import to require

barak007: do not resolve

PR #773, opened 2019-09-09
Add the ability to generate a namespace manifest (for standalone builds)

tomrav:
Adds support in CLI / Webpack-plugin.
Requires an entry in the `package.json`.


tomrav:
Abandoned this attempt of extended backwards compatibility.

PR #781, opened 2019-09-15
Set `@stylable/runtime` as dependancy of `@stylable/module-utils`

NitayRabi:
Module utils now depends on structure available only on `runtime@alpha`
(within module-factory) 

Case - 
Tried updating @ui-autotools independently and it failed to find path `@stylable/runtime/cjs/index-legacy`

### DIFF
--- a/packages/module-utils/package.json
+++ b/packages/module-utils/package.json
@@ -11,7 +11,8 @@
     "prepack": "yarn build"
   },
   "dependencies": {
-    "@stylable/core": "^2.3.1-alpha.0"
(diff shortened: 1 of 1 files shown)


barak007:
This package is only responsible for generating the source code of the JS module it does not requires anything directly from the runtime. The `@stylable/runtime/cjs/index-legacy` reference is used when consumer is evaluating the JS module. I believe this error happens due to mixture of legacy packages in the project. Adding runtime here will not resolve the core issue.

Issue #779, opened 2019-09-11
additional argument warning for custom value


idoros:
Using a [custom value](https://stylable.io/docs/references/variables#advanced-variable-types) which **works as expected** still warns about the second argument:

```css
:import {
    -st-from: "./color-custom-value";
    -st-default: hue;
}
:vars {
    mainColor: hue('170');
}
.root {
    color: value(mainColor, dark);
    /*
Stylable: file.st.css:9:6: value function accepts only a single argument: "value(mainColor, dark)"
    */
}
```

tomrav:
Resolved in #858

PR #809, opened 2019-10-03
Diagnostics: Remove extra warning on value called with several arguments

skad0:
Resolves https://github.com/wix/stylable/issues/779 because of https://stylable.io/docs/references/variables#advanced-variable-types


idoros:
@skad0 I'm not sure the args amount diagnostic should be completely removed, maybe just extended to 2 arguments, ideally only if the value passed in supports it.

skad0:
@idoros 
following by code (and also I reproduced example) it can accept as much arguments as it possible :) 
```
:vars {
    colors: stMap(
        bg stMap(
          more red
        ),
        text red
    );
}

.select {
    background-color: value(colors, bg, more);
}
```

But by checking arguments type - good thing, will check it

tomrav:
Just merged #858 which should resolve the issue, thanks for the PR - it moved us in the right direction 👍

PR #887, opened 2019-11-24
jest: specify cache key method

### LINKED ISSUE #875: @stylable/jest: not declaring getCacheKey causes stale cached results
`@stylable/jest` uses `@stylable/module-utils` to create the module for jest to run.

It sets the runtime path as `require.resolve('@stylable/runtime')`, which is correct, since we want those stylesheets to use the same `@stylable/runtime` the jest integration brings (deduped or not).

The result of `require.resolve('@stylable/runtime')` might change on version upgrades, as packages get deduped. The `node_modules` structure might change, yet `jest` will keep the same transformer result in its cache, with the outdated runtime path.

Suggested solution: declare `getCacheKey` exported next to `process`, and add the "current" resolved runtime path to the key (alongside the defaults).

Ref: https://github.com/facebook/jest/blob/v24.9.0/packages/jest-transform/src/ScriptTransformer.ts#L101

AviVahl:
so that if the runtime location is changed locally, the module will be re-created.

fixes #875

### DIFF
--- a/packages/jest/src/jest.ts
+++ b/packages/jest/src/jest.ts
@@ -2,6 +2,8 @@ import { StylableConfig } from '@stylable/core';
 import { Options, stylableModuleFactory } from '@stylable/module-utils';
 import fs from 'fs';
 
+const stylableRuntimePath = require.resolve('@stylable/runtime');
+
 export function processFactory(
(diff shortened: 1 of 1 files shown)


tomrav:
it's a good start but we need to include all deps as well. -Barak.

PR #924, opened 2019-12-27
Add optional "noCollisionNamespace" namespace resolver

barak007:
This option will enable shorter and cleaner class names in development. and will throw Error on duplicate namespace usage by different files.


tomrav:
We should add a test to check for the new collision functionality.

idoros:
So 2 `button.st.css` files will cause an error?, what is the default recommended configuration required to use identical named files?

tomrav:
> So 2 `button.st.css` files will cause an error?, what is the default recommended configuration required to use identical named files?

An error will be generated. This option will be set to `false` by default. It is intended for smaller projects that keep strict control on their stylable files management, gaining in return a cleaner, more simple output code.

idoros:
If you already have a way to opt-in handle the namespace resolution in both the CLI and webpack plugin, then why add another option? Wouldn't it be simpler to offer this naming function as an external util that can be used externally?

tomrav:
Looking at the code in the change, it's not a config option per se, but a ready-made namespace resolver function to be used in your configuration..

packages/core/src/resolve-namespace-factories.ts
```diff
@@ -0,0 +1,49 @@
...
+            prefix +
+            namespace +
+            hash.v3(
+                hashSalt + config.name + '@' + normalizeVersion(config.version) + '/' + fromRoot
+            )
+        );
+    };
+}
+
+export function noCollisionNamespace({
```
tomrav, line 30: Please add a test for this.

PR #936, opened 2020-01-13
add @st-import syntax as an alternative to :import

barak007:
Since our `:import` is very verbose we can offer simpler and "valid" css syntax to imports.
it's basically esm imports but replacing the Curly brackets with Brackets  and no support for `* as` imports.
It's fully compatible with our current model so it can be a minor change.

New syntax examples: 
```css
@st-import Button from "./button.st.css";
@st-import Button, [icon, text] from "./button.st.css";
@st-import [icon, text as myText] from "./button.st.css";
```


tomrav:
We are missing language service completions for this feature.

tomrav:
We're still missing documentation and LSP integration on this one.

PeterShershov:
@tomrav after testing this on the current stylable-intelligence version with the example in the description, I got this output after formatting the file
image

tomrav:
> @tomrav after testing this on the current stylable-intelligence version with the example in the description, I got this output after formatting the file

How did you test this out? Once this feature is released a new version of intelligence will need to be published as well, I have some work on that locally but it's blocked until this gets merged and published.

tomrav:
I believe the only thing missing now, is creating exceptions for `@st-imports` when formatting the document due to the breakage described above.

tomrav:
Added exception for st-imports when formatting the stylesheet.
In the future we can add a custom parser/printer for those statements alone

packages/core/src/stylable-processor.ts
```diff
@@ -769,6 +783,36 @@ export class StylableProcessor {
...
+            this.diagnostics.error(atRule, processorWarnings.ST_IMPORT_STAR());
+        } else if (imports) {
+            importObj.defaultExport = imports.defaultName || '';
+            setImportObjectFrom(imports.from || '', this.dirContext, importObj);
+            importObj.named = imports.named || {};
+            if (!imports.from) {
+                this.diagnostics.error(atRule, processorWarnings.ST_IMPORT_EMPTY_FROM());
+            }
+        } else {
+            this.diagnostics.error(atRule, processorWarnings.INVALID_ST_IMPORT_FORMAT());
```
idoros, line 809: Worth passing toki errors into the diagnostics for a more informative error

packages/stylable.io/docs/references/imports.md
```diff
@@ -6,39 +6,95 @@ layout: docs
 
 **Stylable** enables you to import other stylesheets and modules in a way that is similar to [JS Imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). You can then use the stylesheet or module as it's been defined, or just one or more of its named values, in your own **Stylable** stylesheet.
 
-You use the **Stylable** syntax beginning with `-st-` for the `:import` config:
+## Import syntax
```
idoros, line 9: Generally I think we should treat the old verbose syntax alive as `legacy` and refer to it as such instead of `old`

packages/stylable.io/docs/references/imports.md
```diff
@@ -6,39 +6,95 @@ layout: docs
...
+**Stylable** supports two types of import syntax that can be used, the two are identical in their capabilities.
```
idoros, line 10: Maybe arrange into:

- import definitions separated from specific syntax: `from`, `default` & `named`
- supported modules short explanation with links for more details
- syntax: new and legacy (keep terminology identical between them)

packages/stylable.io/docs/references/imports.md
```diff
@@ -6,39 +6,95 @@ layout: docs
...
+> * Multiple imports may conflict in their used symbol names; the last one in the file wins.
 
 ## Imports - basic usage
 
 Here are some examples of how you can use imports in your **Stylable** stylesheet.
 
 ### Import the default export of a local reference stylesheet for use in the scoped stylesheet
 
-Import the `toggle-button.css` stylesheet from a local location. Assign the name `ToggleButton` to the default export of that stylesheet for use in this scoped stylesheet. 
+Import the `button.st.css` stylesheet from a local location. Assign the name `Button` to the default export of that stylesheet for use in this scoped stylesheet.
```
idoros, line 46: suggested rewrite:

Import the `button.st.css` stylesheet from a relative location, and assign the local name `Button` to It's default export root element. 

> **NOTE:**
> Stylesheet default export is always it's root element which is treated as a component and should be named with a capital letter, while the named exports represent other internal parts and definitions (maybe link to "#Import named parts from a local stylesheet")

packages/stylable.io/docs/references/imports.md
```diff
@@ -6,39 +6,95 @@ layout: docs
...
+/* comp.st.css - old syntax */
+:import {
+    -st-from: './button.st.css';
+    -st-named: label, icon, --bgColor;
+}
+```
+
+```js
+/* ES6 "equivalent" */
+import { label, icon, --bgColor } from './button.st.css';
```
idoros, line 97: I think that the es6 equivalents examples can be removed because they might suggest that importing an `st.css` file might resolve in that way - which it doesn't.

packages/stylable.io/docs/references/imports.md
```diff
@@ -78,15 +144,34 @@ The values `gridMixin` and `tooltipMixin` are imported from the local JavaScript
...
+    -st-named: keyframes(slideX, slideY);
+}
+```
+
+You can read more about keyframes behavior [here](./keyframes.md).
 
-When importing another stylesheet, the default import represents the root of the stylesheet and is generally treated as a component, and named imports represent vars and classes.
 
-## Usage
+## Importing specific symbols
```
idoros, line 169: heading is a bit misleading because the content links might contain `import` usages, but it doesn't specifically link to them.

idoros (commented):
documentation review

PR #979, opened 2020-02-10
@st-scope and CSS service quality of life improvements

tomrav:
#### Add support for the following within `@st-scope`
- Stylable and native CSS completions
- Native CSS diagnostics
- Native CSS color info (vscode color squares) and presentation (editing color picker) 

#### Media query diagnostics fix
- Using a Stylable value in a media query parameter no longer issues a warning or break css completions


tomrav:
What do you mean by limiting to a single entity? That's a limitation that exists in the `@st-scope` as it accepts a single symbol (class or element)

barak007:
I know but in the future the quote approach will cover every case. Not really important you can decide what to take.

PR #987, opened 2020-02-16
core: allow consumers to specify module resolver

AviVahl:
- not just options for `enhanced-resolve`.
- `webpack-plugin` uses it to create the default resolver with the wrapped `inputFileSystem`.
- fixes 3rd party resolution when `tsconfig-paths-webpack-plugin` is used, as it requires `readJson`.


AviVahl:
Approved by @barak007.

PR #1105, opened 2020-04-16
Asset replacement process only once

barak007:


packages/webpack-plugin/src/stylable-webpack-plugin.ts
```diff
@@ -177,6 +151,32 @@ export class StylableWebpackPlugin {
         });
         this.injectStylableCSSOptimizer(compiler);
     }
+    private processStylableModuleAssets(
+        compilation: webpack.compilation.Compilation,
+        module: StylableModule
+    ) {
+        const rootContext = (compilation as any).options.context;
```
AviVahl, line 158: `compilation.context`

packages/webpack-plugin/src/stylable-webpack-plugin.ts
```diff
@@ -177,6 +151,32 @@ export class StylableWebpackPlugin {
...
+        const replacements: WebpackAssetModule[] = [];
+        const moduleDir = dirname(module.resource);
+        const onUrl = (node: any) => {
```
AviVahl, line 161: `const onUrl: OnUrlCallback = (node) => {`
which will then show you we need to protect against `node.url` being `undefined` before calling isAsset

packages/webpack-plugin/src/stylable-webpack-plugin.ts
barak007: I think I saw a difference but will change and see what is going on.

PR #1114, opened 2020-04-23
docs(migration-v3): add doc with some notes about migrating to v3

argshook:
This PR adds a short migration guide to hopefully help those upgrading from stylable 1 to to stylable 2/3

it is quite specific to React but imo is much better than having nothing.

### DIFF
--- a/packages/stylable.io/docs/guides/migration-v3.md
+++ b/packages/stylable.io/docs/guides/migration-v3.md
@@ -1,0 +1,138 @@
+---
+id: guides/migration-v3
+title: Migrating to Stylable v3
+layout: docs
+---
+
+This guide is intended to help migrate stylable version 1 or 2 to stylable version 3.
+It is mainly geared towards stylable integration within React.
+
+# Suggested steps of migration
+
+Follow these steps for a smooth transition. Each step is explained in
+more detail below.
+
+1. Update dependencies
+1. Update `.st.css` file imports
+1. Update usage in React components
+1. Update tests
+
+## Update dependencies
+
+Ensure you have v3 stylable dependencies available. You may need to
+update your `package.json` or ensure that other dependencies bring you
+v3 stylable:
+
+- `@stylable/cli`
+- `@stylable/core`
+- `@stylable/runtime`
+- `@stylable/node`
+- `@stylable/webpack-plugin`
+
+
+> Note: all stylable packages in v3 are scoped under `@stylable` namespace. 
+> if you have dependency like `stylable` (without namespace),
+> it is a different one and should be changed to namespaced version.
+
+## Update `.st.css` file imports
+
+Prior to v3 all `.st.css` files would export `style` function. In v3
+this has changed: `.st.css` files now export an object.
+
+List of all exported keys:
+
+```
+import {
+  classes,
+  cssStates,
+  keyframes,
+  namespace,
+  st,
+  stVars,
+  style,
+  vars,
+} from './style.st.css';
+```
+
+This means that all imports of `.st.css` files have to be changed, for example:
+
+```diff
+-import style from './Component.st.css';
++import { style, classes, /* ... */ } from './Component.st.css';
+```
+
+However, most often `{ style, classes }` is enough:
+
+`import { style, classes } from './Component.st.css';`
+
+## Update usage in React components
+
+Once `.st.css` imports are updated, React components should be updated too:
+
+```diff
+-<div {...style('root', states, props)} /> />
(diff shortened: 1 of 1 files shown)


packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,138 @@
...
+> Note: find more details and examples in React integration guide https://stylable.io/docs/getting-started/react-integration
+
+## Update tests
+
+If you were using `@stylable/dom-test-kit`, it's usage has slightly
+changd in v3:
+
+```diff
+import { StylableDOMUtil } from '@stylable/dom-test-kit';
+-import * as style from './Component.st.css';
```
argshook, line 125:
```suggestion
-import style from './Component.st.css';
```

packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,138 @@
...
+/>
+```
+
+Stylable v3 usage is like so:
+
+```
+className={style(classes.root, states, this.props.className)}
+```
+
+There is no more props sprad anymore and stylable requires only
```
argshook, line 104:
```suggestion
There is no more props spread anymore and stylable requires only
```

packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,138 @@
...
+changd in v3:
+
+```diff
+import { StylableDOMUtil } from '@stylable/dom-test-kit';
+-import * as style from './Component.st.css';
++import * as style from './Component.st.css';
+const stylableDOMUtil = new StylableDOMUtil(style);
+```
+
+Just as before, `StyalbleDOMUtil` accepts an argument, which is all that
```
argshook, line 130:
```suggestion
Just as before, `StylableDOMUtil` accepts an argument, which is all that
```

packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,138 @@
...
+Once `.st.css` imports are updated, React components should be updated too:
+
+```diff
+-<div {...style('root', states, props)} /> />
++<div className={style(classes.root, states, props.className)} />
+```
+
+There is a subtle but very important nuance in this change.
+
+Stylable v1 and v2 used spread pattern. It would take the output of
```
barak007: only v1 uses the spread

packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,138 @@
...
+`style('root', states, this.props)` function and spread it on component.
+
+This way one or more props would be applied to the component. Thus, code that looks like this:
+
+```
+<div {...style('root', {}, { className: 'additional-class', dataHook: 'test', hello: 'world' })} />
```
barak007: only explicit `data-`  and className will be copied other props like `dataHook` (camelCase) and `hello` will not

packages/stylable.io/docs/guides/migration-v3.md
argshook: hey, thanks for reviewing! However, i believe you're looking at older PR version, @tomrav has already helped me with most of the mistakes

packages/stylable.io/docs/guides/migration-v3.md
```diff
@@ -0,0 +1,165 @@
...
+  export = stylesheet;
+}
+```
+
+This way TypeScript compiler will help moving through most of the
+required changes and provide typings for other Stylable use cases.
+
+## Update `.st.css` file imports
+
+Prior to v3 all `.st.css` files would export `style` function. In v3
```
AviVahl, line 54:
```suggestion
Prior to v2 all `.st.css` files would export `style` function. In v2
```

packages/stylable.io/docs/guides/migration-v3.md
tomrav: I think this is the last thing that needs fixing, then I can merge.

argshook: this is a comment from on an outdated commit, it has been fixed

tomrav: I still see `dataHook: 'test'` being used in the example, it should use an example starting with `data-`.

argshook: ahh, i missed it, fixed now! thanks for your patience :)

PR #1153, opened 2020-05-12
chore(deps): upgrade to latest stable

AviVahl:
- regenerate lock file
- fixed type issues due to updated upstream postcss walkRules type


barak007:
any progress to remove the SRule is good 10x

AviVahl:
Holding this one until typescript-eslint fixes a regression that was introduced in 2.32.

AviVahl:
new typescript-eslint release is out and fixes regression.

typescript@3.9 is also out, and fails with stylable-runtime's hacky custom bundling script. not really surprising. FYI @barak007

PR #1164, opened 2020-05-17
fix issue with mixins and formatters as arguments

barak007:
When using mixins and formatters as arguments the formatter arguments turned in to nulls
```css
.root {
 -st-mixin: Mixin(param formatter(a, b, c))
}
```
the results was
```css
.root {
  color: anullbnullc;
}
```


packages/core/src/stylable-value-parsers.ts
```diff
@@ -297,7 +297,9 @@ export const strategies = {
 
 function stringifyParam(nodes: any) {
     return postcssValueParser.stringify(nodes, (n: any) => {
```
tomrav, line 299: We can check if we can use actual typings here instead of `any`

PR #1174, opened 2020-05-20
fix: timed cache without timer flakiness

AviVahl:
the first test in `timed-cache.spec.ts` depends on Date.now() being 1ms exact.

allow it to clean if diff from last call === timeout (which is also correct), hoping this would fix the flakiness.


AviVahl:
both PR and Branch builds passed once. re-running to verify.

AviVahl:
Passed again. Running branch and pr builds one more time...

AviVahl:
Passed third time. Running again.

AviVahl:
Passed again. One last time...

PR #1176, opened 2020-05-21
LSP: add document formatting providers

tomrav:
Implemented `onDocumentFormatting` and `onDocumentRangeFormatting` language service methods.

Hardcoded the option `space_around_selector_separator` to `true` as `@custom-selector` can break without it.


packages/language-service/src/lib/service.ts
```diff
@@ -233,9 +236,68 @@ export class StylableLanguageService {
...
+
+        return [];
+    }
+
+    public onDocumentRangeFormatting(
+        filePath: string,
+        offset: { start: number; end: number },
+        options: { tabSize: number; insertSpaces: boolean }
+    ) {
+        const stylableFile = this.readStylableFile(filePath);
```
AviVahl, line 268: readStylableFile ignores untitled, which is not ideal for formatting. consider making this provider a bit more flexible, as it only needs the content.

Issue #908, opened 2019-12-16
Webpack: Cannot load assets from node modules


simast:
Is there a feature in Stylable webpack plugin that allows loading assets from node modules? In css-loader you can use tilde (`~`) character for this.

Would it be possible to add support for tilde to Stylable webpack plugin as well? As something like this currently does not work as expected:

```css
.root {
  background-image: url('~wix-ui-icons-common/ArrowRight.svg');
}
```

tomrav:
I'm currently looking at this behavior, trying to figure out what's our best course of action.

tomrav:
PR #929 provides a workaround to this issue and will allow you to properly mix in classes pointing to assets in the same package. This fix will go into the latest version, and will probably not be back ported to `v1`.

Example:
```
/* wix-ui-icons-common/asset-mixins.st.css */

.arrowRight {
  background-image: url('./ArrowRight.svg');
}
```

```
/* my-project/comp.st.css */
:import {
  -st-from: 'wix-ui-icons-common/asset-mixins.st.css';
  -st-named: arrowRight;
}

.part {
  -st-mixin: arrowRight;
}
```

Output:
```
/* my-project/comp.st.css */
.part {
    background-image: url('./node_modules/wix-ui-icons-common/ArrowRight.svg');
}
```

I feel like we should still discuss the option of supporting the `~` operator, but am not sure on the timetables for it.

simast:
Thanks Tom,

But I don't think this is a valid workaround for us right now. There is just too many ways to reference assets in CSS and exposing mixins for each of them is not a solution (e.g. some users will want to use asset as `mask: url(...)`, `content: url(...)` or `list-style-image: url(...)`, not just as background image). 

I think a proper implementation of `~` in the future would be perfect.

tomrav:
I see where you're coming from, and it makes sense to me.

I think the question remains if the correct solution for Stylable would be to use a `~` as suggested, or instead resolve urls in Stylable stylesheets the same way node resolution works (`package/asset.png` for 3rd party, `./asset.png` for local).

tomrav:
After giving this issue further discussion, the solution we're are aiming for at the moment would be to add the `~` capability on url assets as suggested in the original post. 

This is the least breaking change, and in future versions we may want to introduce different modes of behavior.

Examples showing the suggested stylable resolution and the resulting translated node request:
```css
/* relative requests */
background: url(assets/asset.png)    /* resolves as: "./assets/asset.png" */
[...]

PR #1213, opened 2020-06-16
Support '~' url syntax for 3rd party assets

tomrav:
### `core`

Support new syntax (`~`) for importing url assets from a 3rd party
```css
.root {
  background: url(~external-package/kitty-bg.png);
}
```

Resolves issue #908


packages/core/src/stylable-assets.ts
```diff
@@ -28,7 +28,7 @@ export function isAsset(url: string) {
 export function makeAbsolute(resourcePath: string, rootContext: string, moduleContext: string) {
     const isAbs = path.isAbsolute(resourcePath);
     let abs: string;
-    if (isExternal(resourcePath)) {
+    if (isExternal(resourcePath) || (!isAbs && !resourcePath.startsWith('.'))) {
```
barak007, line 31: why the || addition

PR #1223, opened 2020-06-23
security: replace url-regex with is-url-superb

AviVahl:
`url-regex` has a security vulnerability.

`is-url-superb` uses native `URL` api to verify text is a valid url. much cleaner.

ref:
https://github.com/advisories/GHSA-v4rh-8p82-6h5w
https://github.com/kevva/url-regex/issues/70


barak007:
LGTM but I saw the code and it can will swallow the reference error if URL is not available and just return false. I think it should throw the error.

PR #1233, opened 2020-06-29
infra: set importsNotUsedAsValues to "error"

AviVahl:
- forces us to use `import type` when the imported symbol is not used as value.
- the `import type` statements bump min `typescript` to 3.8+. might want to consider running `downlevel-dts` on @stylable/runtime.


AviVahl:
Please do not merge before we discuss `downlevel-dts`.

PR #1238, opened 2020-06-30
Experimental-loader: add exportsOnly mode for transform loader

barak007:
Allow SSR builds to not extract css and get only the runtime stylesheet exported

### DIFF
--- a/packages/experimental-loader/README.md
+++ b/packages/experimental-loader/README.md
@@ -56,7 +56,7 @@ module.exports = {
 interface LoaderOptions {
   resolveNamespace?(namespace: string, filePath: string): string;
   filterUrls?(url: string, ctx: loader.LoaderContext): boolean;
-  exportsOnly: boolean;
+  exportsOnly?: boolean;
 }
 ```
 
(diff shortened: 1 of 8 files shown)


packages/experimental-loader/README.md
```diff
@@ -56,13 +56,28 @@ module.exports = {
...
 
 | Option             | Description                                   |
 | ------------------ | --------------------------------------------- |
 | `resolveNamespace` | override default stylesheet namespace process |
 | `filterUrls`       | filter urls from webpack process              |
+| `exportsOnly`      | only export the runtime stylesheet            |
+
+## SSR (exportsOnly)
+
+Sometimes for server pre-rendering build, you don't want to extract css for stylable files and only export the runtime stylesheet. For that case you only need the `transform` loader and enable `exportsOnly` mode.
```
tomrav, line 71:
```suggestion
When building Stylable for consumption in a server-side renderer build, you may want to extract only the exports of the runtime stylesheets and not the content of their CSS. In such a case you would only require the `transform` loader and to enable the `exportsOnly` mode.
```

packages/experimental-loader/README.md
```diff
@@ -56,13 +56,28 @@ module.exports = {
 interface LoaderOptions {
   resolveNamespace?(namespace: string, filePath: string): string;
   filterUrls?(url: string, ctx: loader.LoaderContext): boolean;
+  exportsOnly: boolean;
```
tomrav, line 59:
```suggestion
  exportsOnly?: boolean;
```

PR #1276, opened 2020-07-21
Add support for any selector inside @st-scope

barak007:
Allow any selector in @st-scope params, this is kind of breaking change due to the fact that you can't use `Symbols` in the scoping selector like: 

```css
@st-scope root {} /* this was working previously */
``` 
Now you must add the class `.` if it's a local or imported class.

Because this feature is not widely used maybe we can get away with minor change?


tomrav:
> Because this feature is not widely used maybe we can get away with minor change?

I don't think this statement is necessarily the case. Most projects that come to mind make some use of `@st-scope`.
Maybe this needs to be behind a flag with a diagnostic until we release a major?

Issue #1271, opened 2020-07-20
Support named import of @keyframes declaration


PeterShershov:
Allow importing `@keyframes` as named imports from different files

### spinner.st.css
```
@keyframes spin {
    100% {
        transform: value(rotate);
    }
}
```
### other-spinner.st.css
```
-st-import {
    -st-from: '../../my-file.st.css';
    -st-named: spin;
}

.myClass {
    animation: spin 2s linear infinite;
}
```

tomrav:
Having discussed this issue internally we've come away with the following conclusions:
- seeing as we scope the name of keyframes, we definitely want to provide the ability to import and export them
- once we allow such imports, we run across the problem of keyframes and classes/vars sharing the same namespace in stylesheet imports
- our suggested solution is to import keyframes using a specific utility function inside the `-st-named` property

```css
:import {
  -st-import: './index.st.css';
  -st-named: somePart, someVar, keyframes(someKeyframe); 
}
```

In the future, we can expand this format to support additional type imports  (e.g. classes, elements, vars and css vars) for a more explicit importing experience.

In addition, our upcoming import shorthand (`@import [someClass, keyframes(someKeyframe)] from './index.st.css';`) should also play nice with the suggested syntax.

PR #1297, opened 2020-07-29
Fix transform of mixins animation declarations

barak007:
When using css mixins, the `@keyframes` definition is not present in the partial AST and the transformer dose not know how to transform the `animation` declarations. 

This PR fixes that issue, but there is still the responsibility of making sure that the definitions are included in the built css.

Sine the usage of the mixin is only via css it might not include the file in the build. 
There are couple of solutions to that:
1. import the file with the `@keyframes` definition via `js` somewhere in the project
2. Stylable should be aware and not optimize this case.  (related to #906 )

This is still not solving the problem of issue #1271

### DIFF
--- a/packages/core/src/stylable-transformer.ts
+++ b/packages/core/src/stylable-transformer.ts
@@ -419,7 +419,7 @@ export class StylableTransformer {
         });
 
         if (mixinTransform) {
-            // override the keyframes for mixins since they dose not contain the @keyframes definition
+            // override the keyframes for mixins since they do not contain the @keyframes definition
             meta.keyframes.forEach((keyFrameAtRule) => {
                 const name = keyFrameAtRule.params;
                 keyframesExports[name] = {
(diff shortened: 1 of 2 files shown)


packages/core/src/stylable-transformer.ts
```diff
@@ -395,25 +395,39 @@ export class StylableTransformer {
...
-                    node: atRule,
+                    node: keyFrameAtRule,
                 };
             }
-            atRule.params = keyframesExports[name].value;
+            keyFrameAtRule.params = keyframesExports[name].value;
         });
 
+        if (mixinTransform) {
+            // override the keyframes for mixins since they dose not contain the @keyframes definition
```
tomrav, line 422:
```suggestion
            // override the keyframes for mixins since they do not contain the @keyframes definition
```

barak007:
This will be resolved via  #1310

PR #1350, opened 2020-09-07
Allow using JavaScript string values as Stylable import

barak007:
Allow usage of imported JavaScript string values inside Stylable value() function.

```css 
:import {
    -st-from: "../my-js-values.js";
    -st-named: myValue;
}

.myClass {
    color: value(myValue);
}
```


tomrav:
Merging but will need to update docs separately

PR #1364, opened 2020-09-14
Fix Removing stylable modules that are used in JavaScript

barak007:
I failed to reproduce the scenario that cause webpack to split modules in a way that the original reason is not in the same chunks of imported stylable file.

This PR fixes that problem by never removing Stylable files that used in JavaScript imports


AviVahl (approved):

AviVahl:
maybe simplify the two ifs into one now:
```ts
if (!shouldKeep && !m.buildInfo.isImportedByNonStylable) {
    chunk.removeModule(m);
}
```

PR #1365, opened 2020-09-15
feat: support and use postcss@8

AviVahl:
- upgrade postcss and related deps.
- adapted type imports to the new v8 declarations.
- fixed all new lint errors (postcss is better typed now, so `.nodes` is defined correctly in more places, ridding the need for `!`)
- reported 3 parser regressions upstream (when using with `postcss-safe-parser`).


AviVahl:
Work left: 2 failing tests in "global scope reference"

AviVahl:
8.0.4 fixed the at-rule exception, but introduced another regression when semi-colons are missing.
I've fixed two tests with the missing semi-colons, and reported the issue upstream.

packages/cli/test/build.spec.ts
```diff
@@ -128,7 +128,7 @@ describe('build stand alone', () => {
         const fs = createFS({
             '/comp.st.css': `
                 :import {
-                    -st-from: "./missing-file.st.css"
+                    -st-from: "./missing-file.st.css";
```
barak007, line 131: nice catch

packages/core/test/stylable-resolver.spec.ts
```diff
@@ -465,7 +465,7 @@ describe('stylable-resolver', () => {
                     namespace: 'A',
                     content: `
                         :import {
-                            -st-from: "b/index.st.css"
+                            -st-from: "b/index.st.css";
```
barak007, line 468: nice catch

barak007 (commented):
10x Avi. This PR looks fine but I saw that the new version dose not allow to extend ast nodes. I don't know that this mean exactly but we are adding keys to the ast nodes. It might be still ok but I need to check it before merging.

tomrav (approved):
Really nice upgrade

AviVahl:
Approved by @barak007 as well.
Rebasing and will merge once tests pass

PR #1408, opened 2020-10-05
refactor: decouple language-service from vscode

AviVahl:
- remove direct dependency and imports to `vscode-languageserver`, `vscode-languageserver-types`, `vscode-languageserver-protocol`.
- instead, import relevant types from `vscode-css-languageservice`, which reexports the ones it's compatible with.
- remove unused `NotificationTypes` and `LSPTypeHelpers` interfaces. aren't used in `stylable-intelligence` as well.

### DIFF
--- a/packages/language-service/package.json
+++ b/packages/language-service/package.json
@@ -23,7 +23,6 @@
     "postcss": "^8.1.1",
     "postcss-value-parser": "^4.1.0",
     "vscode-css-languageservice": "^4.3.4",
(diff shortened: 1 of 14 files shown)


tomrav (approved):

AviVahl:
Our offline discussion made me wonder whether it should actually depend only on the protocol or the types package. Moving to draft.

AviVahl:
will re-open once we have clearer idea of what dependency/type chain we want here.

PR #1509, opened 2020-11-19
core: add the ability to provide a css parser to Stylable

tomrav:


packages/core/src/parser.ts
```diff
@@ -10,6 +10,8 @@ export function cssObjectToAst(cssObject: CSSObject, sourceFile = '') {
     return processor.process(cssObject, { from: sourceFile, parser: postcssJS });
 }
 
-export function safeParse(css: string, options: ProcessOptions = { from: 'style.css' }): Root {
-    return safeParser(css, options);
+export type CssParser = typeof safeParser;
```
barak007, line 13: Maybe typeof postcss.parse

barak007 (changes requested):
I don't think that we want to pass the parser to the safeParse. We want to inject a parse function and not use the safeParse (only as default for now)

packages/core/src/stylable.ts
```diff
@@ -84,7 +87,9 @@ export class Stylable {
             resolveOptions,
             this.resolveNamespace,
             timedCacheOptions,
-            resolveModule
+            resolveModule,
+            undefined,
+            cssParser
```
barak007, line 92: move the cssParser before the 'test-only' argument.

PR #1621, opened 2021-02-07
feat(core-test-kit): Test helper for matching target selector inline

barak007:
When creating Stylable transform tests usually the ugly part was to find the right selector in the AST and match it.

`expect((results.outputAst!.nodes[0] as postcss.Rule).selector).to.equal('...')`

Now we can write the matching selector inside the css as "magic comment" before the tested selector and we must specify the number of assertions.

```js

expectTransformOutput(
    {
        entry: `/style.st.css`,
        files: {
            '/style.st.css': {
                namespace: 'ns',
                content: `
                /* @expect .ns__root */
                .root {}
            `
        },
    },
    1
);

```


tomrav (approved):
This could benefit from being included in the readme, other than that looks good :+1:

PR #1624, opened 2021-02-08
fix(core): fix false report on scoped elements

barak007:
fix a false report on scoped elements when scoping happens after the element node
Example of false report:
`div.root{}`


barak007:
test pending.

PR #1626, opened 2021-02-10
Expand CLI index generation capabilities

### LINKED ISSUE #1625: Allow the CLI more customization when generating a Stylable index file
For instance: Setting the namespace and exposing named imports. cc: @barak007

barak007:
This PR allows to create more capable generator still not the end goal but in the right direction.

* Allow export named parts
* Allow adding heading or footing to the .st.css output

Things that are missing is the ability to get the meta of a stylesheet inside the generate exports function.

closes #1625


packages/cli/test/fixtures/named-exports-generator.ts
```diff
@@ -0,0 +1,40 @@
+import { Generator as Base, ReExports } from '@stylable/cli';
+
+export class Generator extends Base {
```
tomrav, line 3: Maybe this should be moved to `src` and promoted to a public export?

packages/cli/test/generate-index.spec.ts
```diff
@@ -111,6 +111,52 @@ describe('build index', () => {
...
+            rootDir: resolve('/'),
+            log,
+            generatorPath: require.resolve('./fixtures/named-exports-generator.ts'),
+        });
+
+        const res = fs.readFileSync(resolve('/index.st.css')).toString();
+
+        expect(res.trim()).to.equal(
+            [
+                '@namespace "INDEX";',
```
tomrav, line 149: If we change it to an official generator, should the output be this concise, or further spaced apart?

packages/cli/src/base-generator.ts
```diff
@@ -0,0 +1,132 @@
...
+        ...Object.entries(reExports.stVars).map(symbolMapper),
+        ...Object.entries(reExports.vars).map(symbolMapper),
+        ...Object.entries(reExports.keyframes).map(keyframesSymbolMapper),
+    ].join(', ');
+
+    const usagePart = Object.values(reExports.parts)
+        .map((exportName) => `.root .${exportName}{}`)
+        .join(' ');
+
+    return `:import {-st-from: ${JSON.stringify(from)};-st-default:${reExports.root};${
```
tomrav, line 121: When we introduce `@st-import`, should we update the syntax here?

tomrav (changes requested):
Nice feature, should play nicely with our future ones, such as exports management and import destructuring

packages/cli/src/base-generator.ts
barak007: not really. but we can.

packages/cli/test/generate-index.spec.ts
barak007: we will not expose it as a generator but I will expose the function that create the exports for everying

PR #1630, opened 2021-02-14
Enable CLI diagnostics by default and exit process when present

### LINKED ISSUE #1558: CLI builds should fail when there are errors present in the build
Add ability to break the build when there are error in the cli

### LINKED ISSUE #1595: Enable CLI diagnostics by default
`diagnostics` should be be `true` by default when executing `stc`.

can we do it with 4.x?

barak007:
fix(cli)!: report diagnostics true by default
* allow exit cleanly with ignoreDiagnostics option
BREAKING CHANGE:
* process will now exit with failure when there are any diagnostics

I'm not sure that this is the right setup between `diagnostics` and `ignoreDiagnostics` options. Please take a closer look.

closes #1595
closes #1558


barak007:
fix: #1595 #1558

packages/cli/src/cli.ts
```diff
@@ -192,6 +196,15 @@ build({
     useSourceNamespace: useNamespaceReference,
 });
 
+if (diagnosticsMessages.length) {
+    if (diagnostics) {
+        console.log('[Stylable Diagnostics]\n', diagnosticsMessages.join('\n\n'));
+    }
+    if (!ignoreDiagnostics) {
```
tomrav, line 203: Should we be more flexible in how we treat errors vs. warnings?

packages/cli/test/cli.spec.ts
```diff
@@ -256,4 +256,35 @@ describe('Stylable Cli', () => {
         expect(stderr).equal('');
         expect(stdout).to.contain('I HAVE BEEN REQUIRED');
     });
+
+    describe('Cli diagnostics', () => {
```
tomrav, line 260:
```suggestion
    describe('CLI diagnostics', () => {
```

tomrav (approved):
Looks good, strictness huzzah.

packages/cli/src/cli.ts
barak007: yes.

PR #1636, opened 2021-02-15
Add resolver cache for symbol resolution.

barak007:
Add resolver cache for all imports. this is an opt in so there is no breaking change. 
The ability improve dramatically the performance of stylable transform

### DIFF
--- a/packages/experimental-loader/src/cached-stylable-factory.ts
+++ b/packages/experimental-loader/src/cached-stylable-factory.ts
@@ -1,4 +1,5 @@
 import { Stylable, StylableConfig } from '@stylable/core';
+import decache from 'decache';
 import { Compiler } from 'webpack';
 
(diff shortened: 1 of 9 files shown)


packages/experimental-loader/src/cached-stylable-factory.ts
```diff
@@ -13,7 +14,20 @@ export function getStylable(compiler: Compiler, initialConfig: StylableConfig):
...
-        stylable = Stylable.create(initialConfig);
+        const requireModuleCache = new Set<string>();
+        const requireModule = (id: string) => {
+            requireModuleCache.add(id);
+            return require(id);
+        };
+
+        stylable = Stylable.create({ ...initialConfig, requireModule });
+        stylable.initCache();
+        compiler.hooks.done.tap('StylableLoader stylable.initCache', () => {
```
barak007, line 25: todo: only in watch mode 
initCache on create

PR #1638, opened 2021-02-17
refactor(cli): move cli entry to bin folder

AviVahl:
- for a more common package structure.
- named entrypoint `stc.js`, per cli name.

### DIFF
--- a/packages/cli/bin/stc.js
+++ b/packages/cli/bin/stc.js
@@ -3,10 +3,10 @@
 const { normalize } = require('path');
 
 // Use ts-tools to run typescript from source when running in the context of this mono-repo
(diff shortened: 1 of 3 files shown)


barak007 (commented):
look fine to me. this should have been internal concern and i don't believe that some one tried to use the cli from it's exact path.

Thinking if we should put it in v3 or wait for v4?

AviVahl:
Nobody should reference the cli entrypoint directly, and people usually target node_modules/.bin/stc if they want to find the cli. that entry is linked to the js.

AviVahl:
I don't mind this waiting for v4, even though it could enter v3 just as much.

PR #1658, opened 2021-02-21
fix(core): always resolve non-absolute requests

AviVahl:
mid-solution that ensures all non-absolute requests are resolved using the resolver, while absolute ones are returned untouched.

this allows flows calling the resolvePath with absolute paths to stay fast, while providing full/correct resolution for any non-absolute requests.

this change also has the potential to improve caching/correctness in locations that used the returned result as cache key.


barak007:
If this will work I'm good with this middle

AviVahl:
Passed locally and in CI.

I feel like this is a step forward. It should fix that case we saw, but without the implications forcing the resolver for all paths. It's obvious several flows call `resolvePath`, and even ones that don't directly deal with actual module requests. Until we get those refactored, this solution has better balance by covering more use-cases without a major sacrifice of speed.

PR #1681, opened 2021-03-05
fix(webpack-plugin): fix crash in optimization step when css is broken

barak007:
When using `cssInjection: 'css'` if the css is invalid we break webpack. this PR fixes that.

### DIFF
--- a/packages/e2e-test-kit/src/project-runner.ts
+++ b/packages/e2e-test-kit/src/project-runner.ts
@@ -183,7 +183,33 @@ export class ProjectRunner {
             });
         });
     }
-
+    public waitForRecompile() {
+        let done = false;
+        return new Promise<void>((res) => {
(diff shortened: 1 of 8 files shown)


packages/e2e-test-kit/src/project-runner.ts
```diff
@@ -183,7 +183,33 @@ export class ProjectRunner {
             });
         });
     }
-
+    public waitForRecompile() {
+        let done = false;
+        return new Promise<void>((res) => {
+            this.compiler?.hooks.afterDone.tap('waitForRecompile', () => {
```
AviVahl, line 189: inifinite hooks in watch mode?

barak007: no no no. only in watch mode in tests and only if you activate it!!!

PR #1687, opened 2021-03-07
feat(core): basic transform support for @property

barak007:
Stylable will now transform `@property`

```css
//entry.st.css
@property --radius {
  syntax: "<length>";
  inherits: false;
  initial-value: 0px;
}
```
Outputs
```css
//entry.css
@property --entry-radius {
  syntax: "<length>";
  inherits: false;
  initial-value: 0px;
}
```


idoros:
Another restriction that I would add here is to allow `@property` to only register local variables - not imported variables.

PR #1695, opened 2021-03-09
Docs - mixin page - fix explanation

EgoziE:
It was all in one line.
image


barak007:
10x for the PR. We are in the process of migrating our docs to https://github.com/wixplosives/stylable.io. I will add your Content change there

barak007:
added to stylable.io

PR #1702, opened 2021-03-11
fix(webpack-plugin): inject all exports only when usedExports enabled in dev

### LINKED ISSUE #1698: Stylable & Storybook: Cannot read property 'root' of undefined
I'm trying to implement storybook in a Stylable project and have created a button component to try it out:
```ts
import { st, classes } from './Button.st.css'
...
export class Button extends PureComponent<ButtonProps> {
  ...
  render() {
    ...
    return (
      <button type="button" className={st(classes.root, {priority, size})}>{children}</button>
    )
  }
}
```
```
**Cannot read property 'root' of undefined**
TypeError: Cannot read property 'root' of undefined
    at Button.render (http://localhost:6006/main.10c3d77cc995a0fad71f.bundle.js:522:135)
    at finishClassComponent (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:129354:31)
    at updateClassComponent (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:129304:24)
    at beginWork (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:130942:16)
    at HTMLUnknownElement.callCallback (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:115814:14)
    at Object.invokeGuardedCallbackDev (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:115863:16)
    at invokeGuardedCallback (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:115925:31)
    at beginWork$1 (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:135828:7)
    at performUnitOfWork (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:134640:12)
    at workLoopSync (http://localhost:6006/vendors~main.10c3d77cc995a0fad71f.bundle.js:134571:5)
```

When creating the story, the render function of the button component does not seem to know the classes object? Is there something I'm missing? Do I have to change the webpack configuration of Storybook to make this work?

barak007:
Sine webpack can opt out of removing pure exports (like we use in the loader) in certain configurations, this PR moves all injection points in the JavaScript module to the `StylableRuntimeDependency` apply phase.

This allows better decision point for when to inject the shared runtime and also prepare the ground for optimizing `keyframes` and `vars` exports.

fix #1698

### DIFF
--- a/packages/webpack-plugin/src/loader-utils.ts
+++ b/packages/webpack-plugin/src/loader-utils.ts
@@ -9,6 +9,10 @@ import {
 } from '@stylable/core';
 import { dirname } from 'path';
 
+export function getReplacementToken(token: string) {
+    return '/* INJECT */' + JSON.stringify({ [`__${token}__`]: true });
+}
+
 export function getImports(
     stylable: Stylable,
     meta: StylableMeta,
--- a/packages/webpack-plugin/src/loader-utils.ts
+++ b/packages/webpack-plugin/src/loader-utils.ts
@@ -10,7 +10,7 @@ import {
 import { dirname } from 'path';
 
 export function getReplacementToken(token: string) {
-    return '/* INJECT */' + JSON.stringify({ [`__${token}__`]: true });
+    return `/* INJECT */ {__${token}__:true}`;
 }
 
 export function getImports(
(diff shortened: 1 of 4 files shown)


packages/webpack-plugin/src/loader-utils.ts
```diff
@@ -9,6 +9,10 @@ import {
 } from '@stylable/core';
 import { dirname } from 'path';
 
+export function getReplacementToken(token: string) {
+    return '/* INJECT */' + JSON.stringify({ [`__${token}__`]: true });
```
AviVahl, line 13: consider stringifying the token or directly passing a safe string

PR #1705, opened 2021-03-16
fix(wepback-plugin): Remove outdated mini-css error

yavorsky:
https://github.com/webpack-contrib/mini-css-extract-plugin/pull/703 was merged and released as a part of `1.3.9`.

The PR just removes error which seems like deprecated.

With stop supporting `[name]` for the `filename` option it became problematic to use the plugin for multiple entries, especially with https://github.com/wix-incubator/tpa-style-webpack-plugin.

Using `mini-css` and controlling the output via it resolves our issue.

### DIFF
--- a/packages/webpack-plugin/src/plugin.ts
+++ b/packages/webpack-plugin/src/plugin.ts
@@ -451,9 +451,6 @@ export class StylableWebpackPlugin {
                     }
                 );
             } else if (this.options.cssInjection === 'mini-css') {
-                throw new Error(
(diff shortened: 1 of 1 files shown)


barak007:
10x for the PR. actually this is not all the job needed to be done in order to enable the feature. We actually need to get the `CssModule` in different way and also re-enable the tests for it. I'm going to look at it very soon.

barak007:
#1706

Issue #442, opened 2018-10-18
Use mini-extract-text-plugin for st.css files


ranyitz:
Currently, we are building a [toolkit](https://github.com/wix/yoshi) which is being used by hundreds of projects that use technologies like, [less](https://github.com/webpack-contrib/less-loader) and [sass](https://github.com/webpack-contrib/sass-loader).

One of our [famous component libraries](https://github.com/wix/wix-style-react) started to use stylable to write css for some of its components. The library outputs `st.css` files and its users use our toolkit which has [stylable-webpack-plugin](https://github.com/wix/stylable/tree/master/packages/webpack-plugin) and knows how to parse those files as part of their bundling process.

Our [problem](https://github.com/wix/yoshi/pull/329#issue-192450841) is that stylable outputs a different asset (in opposing to how sass/less works). It requires our users to add this css file in their `html` files (add a link tag to the head).

It could be really nice if stylable would work like `less`/`sass` loaders and join the same pipeline that lets `mini-extract-text-plugin` to output one css file at the end of the process.

tomrav:
This issue was resolved in #1706

PR #1706, opened 2021-03-16
fix(webpack-plugin): enable mini-css-extract-plugin feature

barak007:
This PR enables the mini-css support in the webpack plugin.


barak007:
fix #442

PR #1718, opened 2021-03-24
Fix stylable modules output empty css after optimizing their original css

barak007:
This PR fixes issue where stylable modules output empty css in rebuilds.

When we optimizing the css for a module we set the optimized css output as the original css, then later if this module should not have been needed to be optimized we didn't have the full original css to process. Now we hold a separate object for all of the recomputed values.


.gitignore
```diff
@@ -6,6 +6,7 @@ cjs
 .history
 .DS_Store
 .vcode/settings.json
+**/test/**/temp-dist
```
AviVahl, line 9: this makes it much harder to make parallel happen

packages/e2e-test-kit/src/project-runner.ts
```diff
@@ -29,19 +29,24 @@ export class ProjectRunner {
...
+        watchedDir = ''
     ) {
         const disposeAfterEach: Set<() => void> = new Set();
         if (watch) {
             const projectToCopy = runnerOptions.projectDir;
-            const tempDir = createTempDirectorySync('local-test');
-            tempDir.path = nodeFs.realpathSync(tempDir.path);
-            const projectPath = join(tempDir.path, 'project');
-            disposeAfterEach.add(tempDir.remove);
+            if(watchedDir){
```
AviVahl, line 38: watchedDir is a string. what if the target folder doesn't exist? rmdirSync fails in that case, even wih recursive.

barak007: I think I didn't saw it is because it's true for deep structure and not the first level of existing one.

PR #1723, opened 2021-03-25
Remove !! from entry request

barak007:
`compile-as-entry` should not add !! to the entry request. we want loaders to affect entry also.


AviVahl:
so that was the issue? :o

AviVahl (approved):

PR #1738, opened 2021-04-04
Fix performance regression in lint process

barak007:
fix(eslint-plugin-stylable): performance regression


AviVahl (commented):
we need to ensure this works in IDE mode and invalidates between lint runs

barak007:
checked works.

PR #1739, opened 2021-04-04
fix(webpack-plugin): use postcss strict parse

barak007:
Fix use postcss strict parser in Stylable instance
Fix issue when parse css in optimization step did not report filename when error thrown.


AviVahl:
@barak007 I think  `@stylable/core` should use the strict parser by default. this way, the cli will also get this behavior by default.

PR #1741, opened 2021-04-06
Initial implementation of split css for in chunks

barak007:
Add ability to split css per chunk for cases where the entries will not be included in the same page.
Bring back the `[name]` interpolation feature for the filename option. (for single extract only work if there is one entry).


packages/webpack-plugin/src/plugin-utils.ts
```diff
@@ -400,3 +430,56 @@ export function getStylableModules(
...
+
+    compilation.emitAsset(
+        filename,
+        new compilation.compiler.webpack.sources.RawSource(cssSource, false)
+    );
+
+    return filename;
+}
+
+export function getEntryPointModules(
```
barak007, line 467: less stylable specific

packages/webpack-plugin/src/plugin.ts
```diff
@@ -108,6 +109,11 @@ export interface StylableWebpackPluginOptions {
     unsafeMuteDiagnostics?: {
         DUPLICATE_MODULE_NAMESPACE?: boolean;
     };
+    /**
+     * Set the strategy of how to spit the extracted css
+     * This option is only used when cssInjection is set to 'css'
+     */
+    extractMode?: 'single' | 'chunks';
```
idoros, line 116: maybe entries instead of chunks? 
chunks sounds like it might do more than just extract per top level chunk (entry)

barak007: right

barak007, line 116:
```suggestion
    extractMode?: 'single' | 'entries';
```

packages/webpack-plugin/src/plugin.ts
```diff
@@ -419,51 +426,62 @@ export class StylableWebpackPlugin {
...
-
-                        compilation.entrypoints.forEach((entryPoint) => {
-                            entryPoint.getEntrypointChunk().files.add(cssBundleFilename);
-                        });
-
-                        compilation.emitAsset(
-                            cssBundleFilename,
-                            new sources.RawSource(cssSource, false)
-                        );
+                        if (this.options.extractMode === 'chunks') {
```
barak007, line 451:
```suggestion
                        if (this.options.extractMode === 'entries') {
```

packages/webpack-plugin/test/e2e/projects/multiple-chunks-split/webpack.config.js
```diff
@@ -0,0 +1,21 @@
...
+        a: './src/a',
+        b: './src/b',
+    },
+    context: __dirname,
+    devtool: 'source-map',
+    plugins: [
+        new StylableWebpackPlugin({
+            filename: '[name].[contenthash].css',
+            cssInjection: 'css',
+            extractMode: 'chunks',
```
barak007, line 17:
```suggestion
            extractMode: 'entries',
```

PR #1745, opened 2021-04-07
fix(core): use request for js imports in the resolver

barak007:
Our resolve module for JS import was receiving the "from" field from the import symbol. This created a case where absolute path without extension was provided to the resolver and was ignored. it was affecting the requireModule function provided to Stylable and required it to resolve extensions. this PR fixes that by providing the request instead.

### DIFF
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -13,7 +13,7 @@ export function createDefaultResolver(fileSystem: MinimalFS, resolveOptions: any
         useSyncFileSystemCalls: true,
         cache: false,
         fileSystem,
-        extensions: ['.js', '.mjs', '.ts'],
+        extensions: Array.from(new Set(['.js', ...resolveOptions.extensions])),
     });
 
     return (directoryPath, request): string => {
--- a/packages/core/src/module-resolver.ts
+++ b/packages/core/src/module-resolver.ts
@@ -13,6 +13,7 @@ export function createDefaultResolver(fileSystem: MinimalFS, resolveOptions: any
         useSyncFileSystemCalls: true,
         cache: false,
         fileSystem,
+        extensions: ['.js', '.mjs', '.ts'],
     });
 
     return (directoryPath, request): string => {
--- a/packages/core/src/stylable-resolver.ts
+++ b/packages/core/src/stylable-resolver.ts
@@ -27,6 +27,13 @@ export interface CSSResolve<T extends StylableSymbol = StylableSymbol> {
     meta: StylableMeta;
 }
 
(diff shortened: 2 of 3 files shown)


packages/core/src/module-resolver.ts
```diff
@@ -13,6 +13,7 @@ export function createDefaultResolver(fileSystem: MinimalFS, resolveOptions: any
         useSyncFileSystemCalls: true,
         cache: false,
         fileSystem,
+        extensions: ['.js', '.mjs', '.ts'],
```
AviVahl, line 16: that doesn't seem quite right. `.mjs` is wrong. imports to such files in node must specify extensions. `.ts` is infrastructure specific (should come from a repository configuration).

packages/core/src/stylable-resolver.ts
```diff
@@ -51,7 +51,7 @@ export class StylableResolver {
         protected requireModule: (modulePath: string) => any,
         protected cache?: StylableResolverCache
     ) {}
-    private getModule({ context, from }: Imported): CachedModule {
+    private getModule({ context, from, request }: Imported): CachedModule {
```
AviVahl, line 54: what's the from? why is it used for the cache key?

packages/core/src/module-resolver.ts
barak007: it's not for webpack. I actually saw that webpack default resolve options are not good for us they does not contain any extensions. also we don't use webpack infra to load the modules so we can't accept any configuration there.
the thing with `.mjs` is that it's not a node import it's Stylable I don't have a strong opinion to how we need to support them so no problem to remove it.
for `.ts` I don't have a problem saying that I can resolve them correctly by default. 

WDYT?

AviVahl: remove .mjs.
.ts should come from a configuration, or by registering a custom resolver. we can't **assume** the user registers ts-node during his bundling, or otherwise.

barak007: ok. accepted

packages/core/src/stylable-resolver.ts
barak007: from the the abs path via path.resolve which should be still ok for the cache key.

AviVahl: I'm not sure how it's ok for the cache key:
`path.resolve(context, './aaa/some-file.js') === path.resolve(context, 'aaa/some-file.js')`

barak007: and....it's the same file.

barak007:
This is already solved by #2135

PR #1758, opened 2021-04-14
refactor(core): replace the process of expanding custom selectors

barak007:
`postcss-selector-matchs` is the only dependeny that uses old postcss version

This PR attempt to remove it and inline the process.

It was based on `postcss-custom-selectors` that is using `postcss-selector-parser`.
Many issues found and fixed during the process including one that needs more investigation 

```css
@custom-selector :--c1 div;
div:--c1 {}
```
will output `divdiv`

One remaining issue related to `postcss-selector-parser` found

```css
:x(!){}
```

will explode the parser


barak007:
@ido do we still need this?

idoros:
The code has changed **ALOT** since this was written. None of it exists any more in this form and the only selector parser that we use is `@tokey/css-selector-parser`. Do the tests added to this PR pass? I would take the test cases into a new PR and attempt at making them pass (which they don't atm). Here are the test cases in the new format:

```
const { sheets } = testStylableCore(`
    @custom-selector :--t1 .root > .icon;
    @custom-selector :--t2 div;
    @custom-selector :--t3 .text;
    @custom-selector :--t4 .x, div;
    @custom-selector :--t5  div, .y;
    @custom-selector :--3xl .3xl;

    /* @rule .entry__root > .entry__icon, .entry__class */
    :--t1, .class {}

    /* @rule div.entry__root */
     .root:--t2 {}

    /* @rule div.entry__text */
    :--t3:--t2 {}

    /* @rule .entry__x, div */
    :--t4 {}

    /* @rule divdiv */
    div:--t2{}

    /* @rule div.entry__x, divdiv, .entry__x.entry__y */
    :--t4:--t5 {}

    /* @rule .entry__3xl */
    :--3xl {}

    /* @rule :not(.entry__text):not(.entry__x, div) */
    :not(:--t3):not(:--t4){}
`);

const { meta } = sheets['/entry.st.css'];

shouldReportNoDiagnostics(meta);
```

PR #1784, opened 2021-05-06
Add ability to filter url asset processing in webpack plugin

barak007:
This PR adds an option to configure a URL asset filter.
This should solve the issue with NextJS asset processing pipeline that needs to filter all the CSS assets.


packages/webpack-plugin/src/plugin.ts
```diff
@@ -147,6 +152,7 @@ const defaultOptions = (
         : defaultOptimizations(isProd),
     optimizer: userOptions.optimizer ?? new StylableOptimizer(),
     target: userOptions.target ?? 'modern',
+    assetFilter: userOptions.assetFilter ?? (() => true),
```
tomrav, line 155: you can use the same default filter function you created above

tomrav (approved):

PR #1796, opened 2021-05-12
Test inline expectation

idoros:
This PR takes the concept for inline expectation tests from `scope-selector-vs.spec` and add some more features:

- [x] selector check
- [x] optional declarations (all + order)
- [x] optional message for check
- [x] mixin rules - check generated rules that are not part of the source
- [x] @atrule params
- [x] @atrule nested rules

format:
```css
/**
* Test transformed stylesheets inline expectation comments
 * rule checking (place just before rule)
 *
 * full options:
 * @check(label)[5] .selector {decl: value}
 *
 * basic:
 * @check .selector
 *
 * with declarations (will check full match and order):
 * @check .selector {decl1: value; decl2: value}
 *
 * label for check:
 * @check(label for test) .selector
 *
 * target generated rules (mixin):
 * @check[4] .selector
 *
 * support multi line declarations:
 * @check .selector {
 *     decl1: value;
 *     decl2: value;
 * }
 *
 * support multi checks:
 * @check .selector
 * @check[1] .selector:hover
 *
 * support atrule params (anything between the @atrule and body or semicolon)
 * @check screen and (min-width: 900px)
 */
```

I converted a lot of tests to see that everything is working, but not all.

### DIFF
--- a/packages/core-test-kit/src/inline-expectation.ts
+++ b/packages/core-test-kit/src/inline-expectation.ts
@@ -60,7 +60,6 @@ export function testInlineExpects(
                     }
                 }
             }
-            // const m = p.text.match(/@check(.*)/s);
         }
     });
     // check
--- a/packages/core-test-kit/src/inline-expectation.ts
+++ b/packages/core-test-kit/src/inline-expectation.ts
@@ -1,11 +1,18 @@
 import type * as postcss from 'postcss';
 
 interface RuleCheck {
+    kind: `rule`;
     rule: postcss.Rule;
     msg?: string;
     expectedSelector: string;
     expectedDeclarations: [string, string][];
     declarationCheck: 'full' | 'none';
+}
+interface AtRuleCheck {
+    kind: `atrule`;
+    rule: postcss.AtRule;
+    msg?: string;
+    expectedParams: string;
 }
 
 /**
@@ -36,6 +43,9 @@ interface RuleCheck {
  * support multi checks:
  * @check .selector
  * @check[1] .selector:hover
+ *
+ * support atrule params (anything between the @atrule and body or semicolon)
+ * @check screen and (min-width: 900px)
  */
 export function testInlineExpects(
     result: postcss.Root,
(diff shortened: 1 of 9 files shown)


packages/core-test-kit/src/inline-expectation.ts
```diff
@@ -0,0 +1,131 @@
...
+        // next must be a rule sense mixin can only add rules
+        if (next?.type === `rule`) {
+            currentRule = next;
+            count--;
+        } else {
+            return;
+        }
+    }
+    return currentRule && count === 0 ? currentRule : undefined;
+}
```
barak007, line 131: missing newline. do you use prettier?

barak007 (changes requested):
LGTM. 
just make sure you apply prettier.

packages/core-test-kit/src/inline-expectation.ts
idoros: Used it at some point, might have forgotten to run before last commit

packages/core-test-kit/src/inline-expectation.ts
```diff
@@ -0,0 +1,131 @@
...
+            const checksInput = p.text.split(`@check`);
+            for (const checkInput of checksInput) {
+                if (checkInput.trim()) {
+                    const check = createRuleCheck(rule, checkInput, errors);
+                    if (check) {
+                        checks.push(check);
+                    }
+                }
+            }
+            // const m = p.text.match(/@check(.*)/s);
```
tomrav, line 54: leftover?

tomrav:
Do you think we could move the `/* @check */` statement to after the selector instead? I think it would read better that way.

packages/core-test-kit/src/inline-expectation.ts
idoros: oops

idoros:
> Do you think we could move the /* @check */ statement to after the selector instead?

The reason I kept it up is because rules are added/generated downwards (mixins & dev-extended-warning-rule), so if the comment is below the rule then it might be hard to figure out what rule it was originally applied. Also I would like to add checks for @rules and that might place the comment far away from the opening declaration.

packages/core-test-kit/src/inline-expectation.ts
```diff
@@ -0,0 +1,143 @@
+import type * as postcss from 'postcss';
+
+type RuleCheck = {
```
AviVahl, line 3: interface

idoros: 👍

idoros:
> Maybe the first comment of a rule might work.

In second thought that won't be good as not all at rules have a body

barak007 (approved):

PR #1819, opened 2021-05-23
chore!: move cssdocs to schema-extract

barak007:
This will remove the need for `os` polyfill when bundling `@stylable/core`

This is also breaking because `getCssDocsForSymbol` public API moved, but I think it's not used in any external package.


AviVahl:
Move the dependency as well

PR #1831, opened 2021-05-26
fix(core): fix parse of :is and :where as nested-pseudo-class

barak007:
We should wait before merge this PR and instead prefer https://github.com/css-modules/css-selector-tokenizer/pull/25


barak007:
fixed in css-selector-tokenizer 0.8

PR #1832, opened 2021-05-26
add error when nesting rules

barak007:
```css

.x {
  .y{}
}
```
should emit error

### DIFF
--- a/packages/core/src/stylable-processor.ts
+++ b/packages/core/src/stylable-processor.ts
@@ -155,7 +155,7 @@ export const processorWarnings = {
         return 'st-namespace-reference dose not have any value';
     },
     INVALID_NESTING(child: string, parent: string) {
-        return `invalid nesting of ${child} inside ${parent}`;
+        return `Stylable does not support nesting of rules within rules, found: "${child}" inside "${parent}"`;
     },
 };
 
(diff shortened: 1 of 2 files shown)


packages/core/src/stylable-processor.ts
```diff
@@ -154,6 +154,9 @@ export const processorWarnings = {
     INVALID_NAMESPACE_REFERENCE() {
         return 'st-namespace-reference dose not have any value';
     },
+    INVALID_NESTING(child: string, parent: string) {
+        return `invalid nesting of ${child} inside ${parent}`;
```
tomrav, line 158:
```suggestion
        return `Stylable does not support the nesting of rules within rules, found: "${child}" inside "${parent}"`;
```

tomrav: @idoros thoughts about this error message?

idoros: Currently support?

barak007: no. but it does parse and find things inside but not transform correctly

packages/core/src/stylable-processor.ts
```diff
@@ -154,6 +154,9 @@ export const processorWarnings = {
     INVALID_NAMESPACE_REFERENCE() {
         return 'st-namespace-reference dose not have any value';
     },
+    INVALID_NESTING(child: string, parent: string) {
+        return `Stylable does not support nesting of rules within rules, found: "${child}" inside "${parent}"`;
```
tomrav, line 158:
```suggestion
        return `nesting of rules within rules is not supported, found: "${child}" inside "${parent}"`;
```

PR #1844, opened 2021-06-03
Fix global type definition to set default export as unknown instead of never

tomrav:
Both in `create-stylable-app` templates and `runtime` readme file.

Related doc from `stylable.io` will be fixed in its repo.


barak007:
Why? what did you found?

tomrav:
> Why? what did you found?

The `never` type that we previously used was gladly accepted by the `dom-test-kit` and so didn't serve the purpose we intended for it.

PR #1865, opened 2021-06-16
feat: make stylable app templates less strict

barak007:
Until the `.d.ts` if fully integrated these strict rules are annoying.


AviVahl (approved):

AviVahl:
I'm guessing this breaks some common stylable syntax use case. Could you clarify which?

tomrav:
This does not play nice with exported CSS variables in inline styling

Issue #1870, opened 2021-06-18
States with paramater which is not a valid className do not get namespace optimizations


tomrav:
Source:
```css
.root:state(123) {

```

Expected:
```css
.s1[class~="s1---state-3-123"] {}
```

Actual:
```css
.s1[class~="app1169059893---state-23-123"] {}
```

Implemented in:
https://github.com/wix/stylable/blob/master/packages/optimizer/src/stylable-optimizer.ts#L83

PR #1874, opened 2021-06-20
feat(core): don't transform custom states to attributes

barak007:
It turns out that we can avoid creating `[class~=...]` for custom states that have invalid class characters. 
Only spaces are really invalid and we always replace them with `_`, other characters are escaped by the `stringifySelector` function.

This might need more eyes to make sure we don't break anything.

This feature is related to #1870 and if we go for it resolve the optimization issue.


tomrav:
Let's add a flag (on by default) for the new all className based behavior. 

This will allow us to prevent a rollback in case we missed something.

barak007:
I think we don't want the flag it's a change that we don't want to look back with it. if there is a bug/error we will prefer to fix it asap.

PR #1877, opened 2021-06-21
Fix states runtime typings

tomrav:
Add a missing `undefined` as a possible state value.


barak007:
remove null from type

PR #1879, opened 2021-06-24
New CLI code formatter for Stylable stylesheets

tomrav:
Add a new CLI `stc-format` to perform code formatting outside the IDE and without the stylable-intelligence extension.

New CLI documentation added to the [package readme](https://github.com/wix/stylable/blob/tom/cli-code-format/packages/cli/README.md#usage-stc-format).

> Todo: add issue about adding support for `editorconfig` files


packages/cli/README.md
```diff
@@ -101,6 +101,28 @@ To transform your project stylesheets to target JavaScript modules containing th
 $ stc --srcDir="./src" --outDir="./dist"
 ```
 
+## Usage `stc-format`
+
+After installing `@stylable/cli`, a new `stc-format` command will be available, running `stc-format --help` will provide a brief description for the options available.
```
idoros, line 106: "a new"? sounds like release notes.
also `--help` is pretty standard and is just below in the option list, I don't you need to mention it above

packages/cli/src/code-format.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+log('Starting code formatting process');
+const formatPathStats = nodeFs.statSync(target);
+
+if (formatPathStats.isFile()) {
+    if (target.endsWith('.st.css')) {
+        formatStylesheet(target);
+    } else {
+        throw new Error('cannot format file, not a Stylable stylesheet (.st.css)');
```
idoros, line 173: maybe support normal css files as well?

packages/cli/src/code-format.ts
```diff
@@ -0,0 +1,187 @@
...
+        } else if (itemStat.isDirectory()) {
+            res = res.concat(readDirectoryDeep(currentlFilePath));
+        }
+    }
+
+    return res;
+}
+
+function formatStylesheet(filePath: string) {
+    log('Formatting: ' + filePath);
```
idoros, line 140: not sure how other formatters log, but I would rather have a log only for actually changed files or errors by default.
logging every file with the path and success/not changed message is very verbose and makes it hard to see what actually changed in big projects.

packages/cli/test/code-format-cli.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+        expect(dirContent[stylesheetPath1]).equal(expectedOutput);
+        expect(dirContent['a/' + stylesheetPath2]).equal(expectedOutput);
+        expect(dirContent['a/' + stylesheetPath3]).equal(expectedOutput);
+        expect(dirContent['a/c/' + stylesheetPath4]).equal(expectedOutput);
+    });
+
+    it('should format a specific stylesheet', () => {
+        const stylesheetPath = 'style.st.css';
+
+        populateDirectorySync(tempDir.path, {
```
idoros, line 61: maybe add another file next to it to see that it doesn't take the base folder?

packages/language-service/test/lib/formatting.spec.ts
```diff
@@ -34,4 +34,141 @@ describe('Formatting', () => {
...
+
+            expect(res).to.eql([
+                {
+                    range: createRange(0, 0, 3, 20),
+                    newText: '.root {\n    color: red;\n\n\n    background: green;\n}',
+                },
+            ]);
+        });
+
+        it('selector_separator_newline: false', () => {
```
idoros, line 142: missing the default behavior test

packages/cli/src/code-format.ts
tomrav: It passes through our LSP which sanitizes and sets up all sorts of things. The user would probably be better off using `js-beautify` directly in their project.

packages/cli/src/code-format.ts
tomrav: This sounds reasonable, I agree.

packages/cli/test/code-format-cli.spec.ts
tomrav: Will do.

packages/language-service/test/lib/formatting.spec.ts
tomrav: Will add.

packages/cli/README.md
tomrav: Removing the new statement. 
The help statement is the same as listed above for `stc`, I see no harm in including it.

packages/cli/src/cli.ts
```diff
@@ -9,6 +9,7 @@ import { reportDiagnostics } from './report-diagnostics';
 const { join, resolve } = nodeFs;
 
 const argv = yargs
+    .usage('$0 [options]')
```
barak007, line 12: what's the $0 means isn't it a command?

packages/cli/src/code-format.ts
```diff
@@ -0,0 +1,187 @@
...
+    const fileContent = nodeFs.readFileSync(filePath, 'utf-8');
+
+    const formatting = lsp.formatDocument(
```
barak007, line 143: we need to see if we can not use the `formatDocument` from lsp because it creates bloat. if possible move the function somewhere else and reuse it in lsp.

packages/cli/src/cli.ts
tomrav: it's the name of the executable that was run, so in most cases it will be `stc`

packages/cli/src/code-format.ts
tomrav: We can probably do something like that, not sure if it should block this from going in

tomrav: I'm doing it.

tomrav: Done.

packages/cli/src/code-format.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// execute all require hooks before running the CLI build
+for (const request of requires) {
+    if (request) {
+        require(request);
+    }
+}
+
+function readDirectoryDeep(dirPath: string, fileSuffixFilter = '.st.css') {
```
barak007, line 121: we already have this function directoryDeepChildren can we reuse?

barak007 (commented):
Amazing! I feel so much lighter.

packages/cli/src/code-format.ts
tomrav: It's async which kinda makes everything just a little bit trickier. Think we can avoid it in this case.

barak007 (approved):

packages/cli/test/code-format-cli.spec.ts
```diff
@@ -0,0 +1,169 @@
+import { expect } from 'chai';
+import { createTempDirectory, ITempDirectory } from 'create-temp-directory';
+import { join } from 'path';
+import { loadDirSync, populateDirectorySync, runFormatCliSync } from './test-kit/cli-test-kit';
+
+describe('Stylable Code Format Cli', function () {
+    this.timeout(25000);
```
AviVahl, line 7: lower or even remove

packages/cli/test/code-format-cli.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+            const fileContent = 'NOT STYLABLE CONTENT';
+            populateDirectorySync(tempDir.path, {
+                [filePath]: fileContent,
+            });
+
+            const { stderr, stdout } = runFormatCliSync(['--target', tempDir.path]);
+
+            const dirContent = loadDirSync(tempDir.path);
+            expect(dirContent[filePath]).to.equal(fileContent);
+            expect(stdout).to.include('');
```
AviVahl, line 147: ?

packages/cli/test/code-format-cli.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+            const fileContent = 'NOT STYLABLE CONTENT';
+            populateDirectorySync(tempDir.path, {
+                [filePath]: fileContent,
+            });
+
+            const { stderr, stdout } = runFormatCliSync(['--target', join(tempDir.path, filePath)]);
+
+            const dirContent = loadDirSync(tempDir.path);
+            expect(dirContent[filePath]).to.equal(fileContent);
+            expect(stdout).to.include('');
```
AviVahl, line 165: ?

PR #1882, opened 2021-06-27
feat(webpack-plugin): add utility to internalize projects from node target build in NextJS

barak007:
Not sure if we want to adopt this code, it only works in specific case.
The idea behind the code is good. but I didn't find a way to make the exclude not specific. (help needed) 
If I had a way to tell webpack from the external callback to handle the request with the default behavior it would make more sense.


barak007:
We are going to create a mono-repo for all the external integrations that we don't want as dependency in Stylable.

PR #1883, opened 2021-06-28
fix(webpack-plugin): mini-css-extract-plugin CSSModule build

barak007:
Solve issue with future version of mini-css-extract-plugin

### DIFF
--- a/packages/webpack-plugin/src/mini-css-support.ts
+++ b/packages/webpack-plugin/src/mini-css-support.ts
@@ -46,7 +46,25 @@ export function injectCssModules(
                 }),
                 media: '',
                 sourceMap: null,
+                assets: {},
+                assetsInfo: {},
             });
+
+            try {
+                let syncCheck = false;
+                cssModule.build(
+                    compilation.options,
+                    compilation,
+                    compilation.resolverFactory.get('normal'),
+                    compilation.compiler.inputFileSystem,
+                    () => (syncCheck = true)
+                );
+                if (!syncCheck) {
+                    throw new Error('Expect CssModule build to be sync');
+                }
+            } catch (e) {
+                throw new Error('CssModule build error inside Stylable integration');
+            }
 
             compilation.modules.add(cssModule);
             for (const chunk of chunkGraph.getModuleChunksIterable(module)) {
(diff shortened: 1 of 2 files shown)


packages/webpack-plugin/src/mini-css-support.ts
```diff
@@ -46,8 +46,26 @@ export function injectCssModules(
...
+                    compilation,
+                    compilation.resolverFactory.get('normal'),
+                    compilation.compiler.inputFileSystem,
+                    () => (syncCheck = true)
+                );
+                if (!syncCheck) {
+                    throw new Error('Expect CssModule build to be sync');
+                }
+            } catch (e) {
+                throw new Error('CssModule build error inside Stylable integration');
```
AviVahl, line 66: The original error message and stack is being swallowed

packages/webpack-plugin/src/mini-css-support.ts
```diff
@@ -46,8 +46,26 @@ export function injectCssModules(
...
+                let syncCheck = false;
+                cssModule.build(
+                    compilation.options,
+                    compilation,
+                    compilation.resolverFactory.get('normal'),
+                    compilation.compiler.inputFileSystem,
+                    () => (syncCheck = true)
+                );
+                if (!syncCheck) {
+                    throw new Error('Expect CssModule build to be sync');
```
AviVahl, line 63: Why is this validation needed? Seems oddly specific and is not promised by the API...
Can't we build it asynchronously if needed?

packages/webpack-plugin/src/mini-css-support.ts
barak007: I know. I don't thing they add benefit for the user. sine we are applying the build manually. when this error happens the issue is known. if you think it's a must replay.

packages/webpack-plugin/src/mini-css-support.ts
barak007: I wanted to keep things small. it will cause a greater change the hook that we use is a syncHook and I will need to find a better way to solve it. so for now at least I validate the assumption.

packages/webpack-plugin/src/mini-css-support.ts
AviVahl: It's useful to have those exposed for bug reports.

barak007: So error message is now exposed.

PR #1884, opened 2021-06-29
tokey css selector parser

idoros:
This PR replaces the usage of `css-selector-tokenizer` with `@tokey/css-selector-parser` and on the way deprecated all the utils that were used to handle the old AST.

- deprecate previous api (keep backwards compatibility with deprecation flags and warnings)
    - [ ] `selector-util` to deprecate folder and route usage
    - [ ] deprecate `stylable-utils` that worked with the old selector AST
    - [x] deprecate postcss ast `SRule/SDecl` extensions (set deprecation fields)
    - [ ]` elements` / `resolveSelectorElements`
- new selector utils that support the new parser
    - [x] `parse/stringify`
    - [x] `getStylableData/setStylableData` to cache selector ast on `postcss.Rule/Decl`
    - [x] `walk` selector
- [ ] validations for tokey extended functional selectors

still WIP - need to go through all packages and replace deprecated usage


packages/core/src/stylable-meta.ts
```diff
@@ -76,7 +81,7 @@ export interface StylableDirectives {
     '-st-root'?: boolean;
     '-st-states'?: MappedStates;
     '-st-extends'?: ImportSymbol | ClassSymbol | ElementSymbol;
-    '-st-global'?: SelectorAstNode[];
+    '-st-global'?: SelectorNode[];
```
barak007, line 84: doc as breaking

PR #1893, opened 2021-06-30
fix: remove non es5 code from runtime stylesheet

barak007:
We might need to consider again to validate es5 compatibility of the runtime so we don't add these kind of mistakes.

### DIFF
--- a/packages/runtime/src/runtime.ts
+++ b/packages/runtime/src/runtime.ts
@@ -1,7 +1,8 @@
 /* eslint-disable no-var */
 import type { Host, StateMap, StateValue } from './types';
 
-export function stylesheet(host: Host = {}) {
+export function stylesheet(host?: Host) {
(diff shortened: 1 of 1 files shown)


yavorsky (approved):

idoros:
Can't we transpire to es5?

barak007:
We don't want any advance syntax that will grow in transpilation. runtime should be small and fast.

sahariko:
@barak007 Consider using an [ESLint plugin](https://www.npmjs.com/package/eslint-plugin-es5) or configuring your `parserOptions`/`env` to `es5` to catch these cases in "real-time" in the future.

tomrav:
I'm opening an issue to discuss our chosen solution for this constant battle (we'll discuss next week).

Until then, merging this and releasing a patch version.

PR #1905, opened 2021-07-06
infra(ci): enable cache in setup-node

AviVahl:
it uses the cache action under the hood, but actually caches npm/yarn's cache directory per platform.
hoping that this done right will cause ci to install faster.


barak007:
run test run!

PR #1908, opened 2021-07-07
fix(core): emit validation diagnostics on @st-imports

barak007:
This PR fixes `@st-import` resolution validation errors. 
We did not emitted warnings on them because we couldn't find the "correct declaration".

### DIFF
--- a/packages/core/test/diagnostics.spec.ts
+++ b/packages/core/test/diagnostics.spec.ts
@@ -1345,5 +1345,113 @@ describe('diagnostics: warnings and errors', () => {
                 );
             });
         });
+
+        describe('@st-import', () => {
+            it('should error on unresolved file', () => {
+                const config = {
+                    entry: '/main.st.css',
+                    files: {
+                        '/main.st.css': {
+                            namespace: 'entry',
+                            content: `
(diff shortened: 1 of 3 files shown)


idoros:
Maybe open an issue tagged for next version to change the behavior to error

packages/core/test/diagnostics.spec.ts
```diff
@@ -1345,5 +1345,113 @@ describe('diagnostics: warnings and errors', () => {
...
+                                'Missing',
+                                './imported.st.css'
+                            ),
+                            file: '/main.st.css',
+                        },
+                    ]
+                );
+            });
+
+            it('should error on unresolved named symbol with alias', () => {
```
tomrav, line 1426: maybe add a test case for a missing file with a named/default import

```
|@st-import Comp, [ part ] from "./missing.st.css";|
```

Looks good otherwise.

tomrav (approved):

packages/core/test/diagnostics.spec.ts
barak007: it's the same flow as the first one so no need

PR #1912, opened 2021-07-08
refactor(core-test-kit): reduce semi-duplicate APIs

barak007:
We had too many ways to test. This PR reduce few of them.

I don't know any other package that is using `@stylable/core-test-kit` so I assume we can snick a patch here. (needs double check)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,10 @@ cjs
 .vcode/settings.json
 **/test/**/temp-dist
 
+# alternative test project dists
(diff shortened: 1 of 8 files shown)


tomrav (approved):

idoros:
> I don't know any other package that is using @stylable/core-test-kit

I think you can add `FOR INTERNAL USAGE` in the readme.
Why is it even published?

barak007:
We will make it private.

Issue #1922, opened 2021-07-11
.d.ts incorrectly defines extended -st-states in a single stylesheet


tomrav:
A class that extends a local class with a state defined does not represent this same state in the generated `.d.ts` file.

Example:
```css
.root {}

.extended {
    -st-states: myState;
}

.shouldHaveState {
    -st-extends: extended;
}

```

```ts
/* THIS FILE IS AUTO GENERATED DO NOT MODIFY */
declare const namespace = "test92253200";

type states = {
    "test92253200__extended": { "myState"?: boolean; }; // notice shouldHaveState is not defined here
};

declare const classes: {
    "root": "test92253200__root";
    "extended": "test92253200__extended";
    "shouldHaveState": "test92253200__shouldHaveState";
};

// .d.ts continues ...
```

```tsx
<div className={st(classes.root, className)}>
    { /* the line below errors on missing state definition in .d.ts */ }
    <span className={st(classes.shouldHaveState, { myState: true})} >
        hello, world.
    </span>
</div>
```

Question: once the extended class moves to another stylesheet how should the .d.ts behave?

PR #1933, opened 2021-07-18
fix(module-utils): locally extended states are now emitted in the generated .d.ts

tomrav:
In the example below, `.target` should be represented in the generated `.d.ts` file as having the `myState` definition.

```css
.x {
  -st-states: myState;
}

.target {
  -st-extends: x;
}
```

This does not and should not happen for states that cross stylesheet boundaries (imported).

### To Do:
- [x] - handle overriding local extended states
- [x] - fix source maps for locally extended states

### Fixes:
- #1922


packages/module-utils/test/sourcemap.spec.ts
```diff
@@ -179,6 +179,85 @@ describe('.d.ts source-maps', () => {
...
+
+        const dtsText = generateDTSContent(res);
+        const sourcemapText = generateDTSSourceMap(dtsText, res.meta);
+
+        sourceMapConsumer = await new SourceMapConsumer(sourcemapText);
+        const state1OriginalPosition = sourceMapConsumer.originalPositionFor(
+            getPosition(dtsText, 'sameState"?:') // source mapping starts after the first double quote
+        );
+
+        expect(state1OriginalPosition).to.eql({
```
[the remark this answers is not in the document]
tomrav: added a comment, hope it's clearer now.

packages/module-utils/src/generate-dts-sourcemaps.ts
```diff
@@ -106,20 +106,38 @@ type Position = {
     column: number;
 };
 
+function findDefiningClassName(stateToken: ClassStateToken, entryClassName: ClassSymbol) {
+    let currentClass = entryClassName;
+
+    while (currentClass['-st-states']) {
```
barak007, line 112: use `valueMapping`

packages/module-utils/src/generate-dts.ts
```diff
@@ -1,8 +1,71 @@
...
+        }
+    }
+}
+
+function collectLocalStates(cls: ClassSymbol) {
+    const stateEntries = new Map<string, StateParsedValue | null>();
+    let currentClass: ClassSymbol | undefined = cls;
+
+    while (currentClass) {
+        const stStates = currentClass['-st-states'];
```
barak007, line 36: use valueMapping

packages/module-utils/src/generate-dts.ts
```diff
@@ -1,8 +1,71 @@
...
+
+        if (stStates) {
+            addStatesEntries(stateEntries, stStates);
+        }
+
+        const extendedClass = currentClass['-st-extends'] as
```
barak007, line 42: why do you need the `as ...`

packages/module-utils/test/generate-dts.spec.ts
```diff
@@ -247,6 +276,75 @@ describe('Generate DTS', function () {
...
+                        import { st, classes } from "./test.st.css";
+                        
+                        eq<string>(st(classes.test, { state1: 'hello' }));
+                    `,
+                });
+
+                expect(tk.typecheck('test.ts')).to.equal('');
+            });
+
+            it('should warn about incorrect state value that is overridden through a local extend', () => {
```
barak007, line 296: Why do we warn about this ? seems legit

barak007: I know this will make the types more complicated but we might want to solve it. (related to states per class)

packages/module-utils/test/generate-dts.spec.ts
```diff
@@ -247,6 +276,75 @@ describe('Generate DTS', function () {
...
+                        eq<string>(st(classes.test, { state1: true }));
+                    `,
+                });
+
+                expect(tk.typecheck('test.ts')).to.include(
+                    `Type 'true' is not assignable to type 'string | undefined'`
+                );
+            });
+
+            it('should not warn about the same state that is being overridden through a local extend to the same type', () => {
```
barak007, line 313: same state per class issue

packages/module-utils/test/generate-dts.spec.ts
```diff
@@ -247,6 +276,75 @@ describe('Generate DTS', function () {
...
+                    'test.ts': `
+                        import { eq } from "./test-kit";
+                        import { st, classes } from "./test.st.css";
+                        
+                        eq<string>(st(classes.test, { state1: true }));
+                    `,
+                });
+
+                expect(tk.typecheck('test.ts')).to.include(
+                    `Argument of type '{ state1: boolean; }' is not assignable to parameter of type 'NullableString'`
```
barak007, line 344: I'm thinking we should rename `NullableString` to `NullableStateValue` WDYT?

packages/module-utils/test/generate-dts.spec.ts
```diff
@@ -247,6 +276,75 @@ describe('Generate DTS', function () {
...
+                tk.populate({
+                    'base.st.css': '.part {stStates: state1}',
+                    'test.st.css': `
+                        @st-import [part] from './base.st.css';
+                        .test { -st-extends: part; }`,
+                    'test.ts': `
+                        import { eq } from "./test-kit";
+                        import { st, classes } from "./test.st.css";
+                        
+                        eq<string>(st(classes.test, { state1: true }));
```
barak007, line 339: I'm not sure that states from imports should be visible in the extending stylesheet

packages/module-utils/src/generate-dts.ts
tomrav: 'extendedClass' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

packages/module-utils/test/generate-dts.spec.ts
tomrav: rephrased the `it`, hopes it's more descriptive.

packages/module-utils/test/generate-dts.spec.ts
tomrav: that's not the only place it's used, we can split them to separate strings if we want.

Issue #1148, opened 2019-07-22
-st-from: "<file-path>" should be clickable


AviVahl:
Example:
```css
:import {
    -st-from: "../src/components/style.st.css";
    -st-default: MyStyle;
}
```

when clicking `../src/components/style.st.css`, it should jump to that file.

PR #1961, opened 2021-07-28
fix(language-service): improve definition support for both :import and @st-import

tomrav:
Go-to-definition with imports behavior should be much better:
- `from` paths are now clickable
- `default` imports work more consistently 
- `named` imports work more consistently
- prevent exceptions from being thrown in certain cases 

This is just an improvement, there are still plenty of quirks.

Fixes: 
- #1148

### DIFF
--- a/packages/language-service/src/lib/provider.ts
+++ b/packages/language-service/src/lib/provider.ts
@@ -204,7 +204,7 @@ export class Provider {
                         /**/
                     }
 
-                    let filePath;
+                    let filePath: string | undefined;
 
                     if (resolved && resolved._kind !== 'js') {
                         filePath = resolved.meta.source;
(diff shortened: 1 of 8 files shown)


barak007 (approved):

packages/language-service/src/lib/provider.ts
```diff
@@ -182,26 +197,33 @@ export class Provider {
...
+                    let resolved: CSSResolve | JSResolve | null = null;
                     try {
-                        rslvd = this.stylable.resolver.resolve(symb);
+                        resolved = this.stylable.resolver.resolve(symbol);
                     } catch {
                         /**/
                     }
 
-                    let filePath: string;
+                    let filePath;
```
tzachbon, line 207: It will always be `any`

tzachbon (commented):
LGTM, small questions left

packages/language-service/src/lib/provider.ts
idoros: why not set it as `let filePath = ""`?

tomrav: fixed it to be `string | undefined` which also works.

PR #1962, opened 2021-08-01
fix(core): assure that pseudo-classes warn when no arguments are given

### LINKED ISSUE #1936: Pseudo-class that expects a param should warn when none is given
```css
.root {
  -st-states: myState(string);
}

.root:myState {} /* should warn about missing argument */
```

tzachbon:
should warn when pseudo-class expects params but none was given

Closes #1936

### DIFF
--- a/packages/core/src/pseudo-states.ts
+++ b/packages/core/src/pseudo-states.ts
@@ -22,7 +22,7 @@ export const stateErrors = {
     TOO_MANY_STATE_TYPES: (name: string, types: string[]) =>
         `pseudo-state "${name}(${types.join(', ')})" definition must be of a single type`,
     NO_STATE_ARGUMENT_GIVEN: (name: string, type: string) =>
-        `pseudo-state "${name}" expected argument of type ${type} but got none`,
+        `pseudo-state "${name}" expected argument of type "${type}" but got none`,
     NO_STATE_TYPE_GIVEN: (name: string) =>
         `pseudo-state "${name}" expected a definition of a single type, but received none`,
     TOO_MANY_ARGS_IN_VALIDATOR: (name: string, validator: string, args: string[]) =>
@@ -298,11 +298,9 @@ function resolveStateValue(
     );
 
     if (rule && !node.content && !stateDef.defaultValue) {
-        diagnostics.warn(
-            rule,
-            stateErrors.NO_STATE_ARGUMENT_GIVEN(name, stateDef.type),
-            { word: actualParam }
-        );
+        diagnostics.warn(rule, stateErrors.NO_STATE_ARGUMENT_GIVEN(name, stateDef.type), {
+            word: actualParam,
+        });
     }
 
     const validator = systemValidators[stateDef.type];
--- a/packages/core/src/pseudo-states.ts
+++ b/packages/core/src/pseudo-states.ts
@@ -295,6 +295,14 @@ function resolveStateValue(
         node.content || stateDef.defaultValue
     );
 
+    if (!actualParam && rule) {
+        diagnostics.warn(
+            rule,
+            `pseudo-state "${name}" expects a parameter of type ${stateDef.type} but none was given`,
+            { word: actualParam }
+        );
+    }
+
     const validator = systemValidators[stateDef.type];
 
     let stateParamOutput;
--- a/packages/core/src/pseudo-states.ts
+++ b/packages/core/src/pseudo-states.ts
@@ -21,6 +21,8 @@ export const stateErrors = {
         `pseudo-state "${name}" defined with unknown type: "${type}"`,
     TOO_MANY_STATE_TYPES: (name: string, types: string[]) =>
         `pseudo-state "${name}(${types.join(', ')})" definition must be of a single type`,
+    NO_STATE_ARGUMENT_GIVEN: (name: string, type: string) =>
+        `pseudo-state "${name}" expected argument of type ${type} but got none`,
     NO_STATE_TYPE_GIVEN: (name: string) =>
         `pseudo-state "${name}" expected a definition of a single type, but received none`,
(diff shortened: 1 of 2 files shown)


packages/core/src/pseudo-states.ts
```diff
@@ -295,6 +295,14 @@ function resolveStateValue(
         node.content || stateDef.defaultValue
     );
 
+    if (!actualParam && rule) {
+        diagnostics.warn(
+            rule,
+            `pseudo-state "${name}" expects a parameter of type ${stateDef.type} but none was given`,
```
idoros, line 301: Add the warning (maybe error?) using the "state Errors" also use the same error function in the tests

packages/core/test/pseudo-states.spec.ts
```diff
@@ -626,6 +626,58 @@ describe('pseudo-states', () => {
...
+                            .my-class {
+                                -st-states: state1;
+                            }
+                            .my-class:state1 {}
+                            `,
+                        },
+                    },
+                }
+
+                const res = expectWarningsFromTransform(config, [])
```
idoros, line 674: I would expect from the title of the test that the test would try to call the state, that was defined with no param, with a param

tomrav: Also I would say that generally we try not to test things that should "not happen"

tzachbon: I agree with that, but how would you make sure it won't happen? @tomrav @idoros

packages/core/src/pseudo-states.ts
tzachbon: Not sure I fully understand. Make it an error sounds legit tho.

packages/core/test/pseudo-states.spec.ts
tomrav: In this case I probably wouldn't, and add a regression test only if it pops up somewhere unexpected somewhere down the line

idoros: I believe there are plenty of non param tests that check for diagnostics that would break if this new one would be reported for them.

Also, slightly related, I think it would be good to have a test that non param states that are called with a param also report an error (if non exist already)

tzachbon: I moved it to be a regression test.

barak007 (commented):
Let's go over this together to see the thinking process.

packages/core/test/pseudo-states.spec.ts
```diff
@@ -626,6 +649,35 @@ describe('pseudo-states', () => {
...
+                const config = {
+                    entry: `/entry.st.css`,
+                    files: {
+                        '/entry.st.css': {
+                            namespace: 'entry',
+                            content: `
+                            .my-class {
+                                -st-states: state1(string);
+                            }
+                            |.my-class:state1 {}
```
idoros, line 662: do you need the `|` (pipe) in this test?

packages/core/test/pseudo-states.spec.ts
```diff
@@ -626,6 +649,35 @@ describe('pseudo-states', () => {
...
+                            `,
+                        },
+                    },
+                }
+
+                const res = expectWarningsFromTransform(config, [
+                    {
+                        message: 'pseudo-state "state1" defined with unknown type: "string"',
```
idoros, line 670: use the same `stateErrors,.*` to create the expected diagnostic message

tzachbon: Used new one.

packages/core/test/pseudo-states.spec.ts
tzachbon: Yep, checking where the diagnostic accord.

packages/core/test/pseudo-states.spec.ts
```diff
@@ -626,6 +649,64 @@ describe('pseudo-states', () => {
                 });
             });
 
+            it('should warn when pseudo-class expects params but none was given', () => {
```
barak007, line 652: all these test are also related to the fact that there is no default value. Add this into the test title

barak007 (changes requested):
One last thing left.

packages/core/test/pseudo-states.spec.ts
barak007: What Ido is saying here is use `stateErrors.UNKNOWN_STATE_TYPE` and don't repeat the error this way when the error is changed you don't need to change the test

packages/core/test/pseudo-states.spec.ts
tzachbon: @barak007 done

packages/core/src/pseudo-states.ts
```diff
@@ -295,6 +297,14 @@ function resolveStateValue(
         node.content || stateDef.defaultValue
     );
 
+    if (rule && !node.content && !stateDef.defaultValue) {
+        diagnostics.warn(
+            rule,
+            stateErrors.NO_STATE_ARGUMENT_GIVEN(name, stateDef.type),
+            { word: actualParam }
```
tomrav, line 304: should probably target an existing string

packages/core/test/pseudo-states.spec.ts
```diff
@@ -626,6 +649,64 @@ describe('pseudo-states', () => {
...
+                const config = {
+                    entry: `/entry.st.css`,
+                    files: {
+                        '/entry.st.css': {
+                            namespace: 'entry',
+                            content: `
+                            .my-class {
+                                -st-states: state1(string);
+                            }
+                            |.my-class:state1() {}
```
tomrav, line 691: fix piping and word selection

tomrav: consider moving diagnostics tests to bottom of file

tomrav (commented):
Overall looks good! :+1:

tomrav (approved):

PR #1967, opened 2021-08-03
fix: property definition override conflicting imported symbol

### LINKED ISSUE #1955: @property definition override conflicting imported symbol
### Actual Behavior

[Playground Link](wixplosives.github.io/stylable-playground/#state=%28%27files%21%282%21%27%2540st-import%2520%255B7%255D%2520from%2520%255C%27.3%255C%275*%2540property%252075*6*%2520%2520%2520%25204red5%29%27%7E3%21%2764green8.label18.icon18%27%29%7Eselected%212%278%255Cn0.st.css1%2520%282%27%252Findex03%252Fbutton047%253A%25205%253B*6.root17--x8%29*%2501876543210*_)

### Expected Behavior

We should not allow to redeclare imported css vars 

### Version

@stylable/core@4.5.0

### Relevant log output

_No response_

### Additional Context

_No response_

tzachbon:
Closes #1955

### DIFF
--- a/packages/core/src/stylable-processor.ts
+++ b/packages/core/src/stylable-processor.ts
@@ -300,6 +300,7 @@ export class StylableProcessor {
 
                     break;
                 case 'property':
+                    this.checkRedeclareSymbol(atRule.params, atRule);
                     this.addCSSVarDefinition(atRule);
                     break;
                 case 'st-global-custom-property': {
@@ -643,7 +644,6 @@ export class StylableProcessor {
 
     protected addCSSVarDefinition(node: postcss.Declaration | postcss.AtRule) {
         const varName = node.type === 'atrule' ? node.params : node.prop;
-        this.checkRedeclareSymbol(varName, node);
         this.addCSSVar(varName.trim(), node);
     }
 
--- a/packages/core/src/stylable-processor.ts
+++ b/packages/core/src/stylable-processor.ts
@@ -643,6 +643,7 @@ export class StylableProcessor {
 
     protected addCSSVarDefinition(node: postcss.Declaration | postcss.AtRule) {
         const varName = node.type === 'atrule' ? node.params : node.prop;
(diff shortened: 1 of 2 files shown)


tomrav:
We should probably also document the `@property` behavior on our site under the CSS var entry.

barak007 (approved):
Nice reuse

packages/core/src/stylable-processor.ts
```diff
@@ -300,6 +300,7 @@ export class StylableProcessor {
 
                     break;
                 case 'property':
+                    this.checkRedeclareSymbol(atRule.params, atRule);
```
idoros, line 303: this is checked only for `@property` because although setting and using css var is considered a declaration, it doesn't redeclare a variable if it exist in the scope?

barak007: right

barak007: it's only for the @property

packages/core/test/stylable-transformer/at-property.spec.ts
```diff
@@ -1,6 +1,7 @@
+import { processorWarnings } from '@stylable/core';
+import { expectWarningsFromTransform, generateStylableResult } from '@stylable/core-test-kit';
 import { expect } from 'chai';
 import type * as postcss from 'postcss';
-import { generateStylableResult } from '@stylable/core-test-kit';
 
 describe('@property support', () => {
```
tomrav, line 6: I think all these tests should move to the `css-vars.spec` file, not sure about this as a separate scope.

tomrav (approved):
Looks good :+1:

packages/core/test/stylable-transformer/at-property.spec.ts
tzachbon: Let's discuss it in a separate issue.

PR #1970, opened 2021-08-04
fix(core):  scope selectors inside a functional pseudo-class correctly

### LINKED ISSUE #1968: any none root selectors after the first in a functional pseudo-class is not transformed correctly
### Actual Behavior

[Playground link](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28*%21%27.a%2520%280%2520%2520%2520%2520-st-states%253A%2520b%252Cc%253B0%2900.a%253Ais%257B%253Ab%252C%253Ac%257D%28%29%27%29%7Eselected%21*%27%29*%27%252Findex.st.css0%255Cn%25010*_)

After the first state inside the `:is`, it does not transform properly.
(It resets the anchor to the root)

### Expected Behavior

After each iteration on the list of chunks, it should reset to the last context and not to the root.
https://github.com/wix/stylable/blob/master/packages/core/src/stylable-transformer.ts#L484

### Version

4.5.0

### Relevant log output

_No response_

### Additional Context

_No response_

tzachbon:
Closes #1968

### DIFF
--- a/packages/core/test/pseudo-states.spec.ts
+++ b/packages/core/test/pseudo-states.spec.ts
@@ -2027,34 +2027,6 @@ describe('pseudo-states', () => {
                 });
             });
         });
-
-        describe('state inside selector list', () => {
-            it('should scope multiple states from none root anchor', () => {
-                const result = generateStylableResult({
-                    entry: `/entry.st.css`,
-                    files: {
-                        '/entry.st.css': {
-                            namespace: 'entry',
-                            content: `
-                                .a {
-                                    -st-states: b, c;
-                                }
-
-                                .a:is(:b, :c) {}
-                            `,
-                        },
-                    },
-                });
-
-                expect(
-                    result.meta.diagnostics.reports,
-                    'no diagnostics reported for imported states'
-                ).to.eql([]);
-                expect(result).to.have.styleRules({
-                    1: '.entry__a:is(.entry--b, .entry--c) {}',
-                });
-            });
-        });
     });
 
     describe('diagnostics', () => {
--- a/packages/core/test/pseudo-states.spec.ts
+++ b/packages/core/test/pseudo-states.spec.ts
@@ -2027,6 +2027,34 @@ describe('pseudo-states', () => {
                 });
             });
         });
+
+        describe('state inside selector list', () => {
+            it('should scope multiple states from none root anchor', () => {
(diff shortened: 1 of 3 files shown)


packages/core/test/pseudo-states.spec.ts
```diff
@@ -2027,6 +2027,34 @@ describe('pseudo-states', () => {
                 });
             });
         });
+
+        describe('state inside selector list', () => {
```
idoros, line 2031: I know I said that  maybe `pseudo-states.spec` is a good place for this, but I mislead you. this spec is for stylable custom states. this is simply a transform of functional pseudo-classes. There is no specific spec file for pseudo-classes, so maybe add a describe for `pseudo-classes` in `scoping.spec` and place it under there for now.

packages/core/test/pseudo-states.spec.ts
```diff
@@ -2027,6 +2027,34 @@ describe('pseudo-states', () => {
...
+                        },
+                    },
+                });
+
+                expect(
+                    result.meta.diagnostics.reports,
+                    'no diagnostics reported for imported states'
+                ).to.eql([]);
+                expect(result).to.have.styleRules({
+                    1: '.entry__a:is(.entry--b, .entry--c) {}',
```
idoros, line 2054: use `testInlineExpects` to check for transformation results

packages/core/test/stylable-transformer/scoping.spec.ts
```diff
@@ -799,6 +799,36 @@ describe('Stylable postcss transform (Scoping)', () => {
 
             testInlineExpects(result);
         });
+
+        it('should scope multiple pseudo-elements that extended none root anchor', () => {
```
idoros, line 803: this test is redundant in my opinion. You already checked the transform context target flow in the other test. it is not specifically related to pseudo-elements

idoros (changes requested):

packages/core/test/stylable-transformer/scoping.spec.ts
idoros: the fix is for the transformation flow of pseudo-classes, the only flow that calls `scopeSelectorAst` with a nested context is a functional pseudo-class

idoros (approved):

barak007 (approved):

PR #1971, opened 2021-08-04
feat(core): support unscoped rules with st-scope

### LINKED ISSUE #1049: Make it possible to define CSS rules in global scope without a warning
```
WARNING in Stylable: unscoped element "html" will affect all elements of the same type in the document

> 109 | html { color: red; }
```

For tag selectors the warning isn't even helpful, what else would anyone expect to happen?

It becomes even more annoying when you want to reset user agent styles, define basic typography, or style third-party components. You have to repeat `:global()` over and over again:

```css
:global(html), :global(body), :global(div), :global(span), 50 more { ... }
```

**Suggestion**:
1. Remove the warning completely for tag selectors.
2. Add an option to suppress the warning for class selectors and component root selectors, e.g.:
```css
:import {
    -st-from: './foo.st.css';
    -st-default: Foo;
}

/* warning */
Foo { }

/* no warning */
@st-scope global {
    Foo {}
}
```

`-st-global: '*'` is not a good workaround because it generates a descendant selector, which could affect specificity and performance.

tzachbon:
We want to allow users to declare unscoped rules without warning.

- [x] - Add note in documentation about this feature (https://github.com/wixplosives/stylable.io/pull/26)

Closes #1049

### DIFF
--- a/packages/core/src/stylable-processor.ts
+++ b/packages/core/src/stylable-processor.ts
@@ -133,9 +133,6 @@ export const processorWarnings = {
     NO_KEYFRAMES_IN_ST_SCOPE() {
         return `cannot use "@keyframes" inside of "@st-scope"`;
     },
-    MISSING_SCOPING_PARAM() {
-        return '"@st-scope" missing scoping selector parameter';
-    },
     ILLEGAL_GLOBAL_CSS_VAR(name: string) {
         return `"@st-global-custom-property" received the value "${name}", but it must begin with "--" (double-dash)`;
     },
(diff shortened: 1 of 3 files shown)


barak007:
Nice. 
I have one thing extra to check. 
In the transformer in the place that we transform the @st-scope selector we want to skip this process when there is an empty selector

tzachbon:
> Nice.
> I have one thing extra to check.
> In the transformer in the place that we transform the @st-scope selector we want to skip this process when there is an empty selector

For every st-scope parameter? I guess it should be in a different PR.

barak007:
look at `validateScopes` in `stylable-transformer.ts` we should not try to transform an empty selector

tomrav:
After discussion, we've decided against going for this approach, will open a separate issue about this.

PR #1973, opened 2021-08-05
feat(core): support defining and using global keyframes

### LINKED ISSUE #1559: Add ability to define global @keyframes
Somehow allow to define global keyframes.
Maybe via `@st-global-keyframes`  or `:global(Name)`

tzachbon:
We want to allow to expose and use global keyframes.
Importing global keyframes should be seamless and be resolve like other keyframes.

- [x] - when declaring global keyframe:
```css

@keyframes stGlobal(my-global-keyframes) {
	from {}
	to {}
}

```

- [x] - add section in `stylable.io`. https://github.com/wixplosives/stylable.io/pull/27

Closes #1559

### DIFF
--- a/packages/core/src/stylable-processor.ts
+++ b/packages/core/src/stylable-processor.ts
@@ -1,14 +1,14 @@
-import { murmurhash3_32_gc } from './murmurhash';
 import path from 'path';
 import * as postcss from 'postcss';
 import postcssValueParser from 'postcss-value-parser';
 import { tokenizeImports } from 'toky';
 import { Diagnostics } from './diagnostics';
+import { murmurhash3_32_gc } from './murmurhash';
+import { reservedKeyFrames } from './native-reserved-lists';
 import {
     createSimpleSelectorChecker,
     isChildOfAtRule,
     isCompRoot,
-    isGlobal,
     isNested,
     isRootValid,
     parseSelector,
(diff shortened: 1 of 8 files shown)


packages/core/test/stylable-transformer/global.spec.ts
```diff
@@ -141,4 +146,91 @@ describe('Stylable postcss transform (Global)', () => {
...
+                        `,
+                    },
+                },
+            };
+
+            testInlineExpects(generateStylableRoot(config));
+            expectWarningsFromTransform(config, []);
+        });
+
+        it('should export global keyframe (alias)', () => {
```
barak007, line 204: rename test something with rename global keyframes

barak007 (approved):
Great work!
LGTM.
We only need approve for the `stGlobal` idea. then we need to update the docs.

packages/core/src/stylable-processor.ts
```diff
@@ -136,6 +141,12 @@ export const processorWarnings = {
     MISSING_SCOPING_PARAM() {
         return '"@st-scope" missing scoping selector parameter';
     },
+    MISSING_KEYFRAMES_PARAM() {
```
tomrav, line 144: This should probably be `MISSING_KEYFRAMES_NAME`, with the texts using `name` instead of `param`.

Param originates from the at-rule side of things, and not from the keyframes/animation semantics.

tomrav (approved):
Looks good, great feature to add.

Left one comment about some diagnostics semantics I think should be changed.

PR #2013, opened 2021-08-16
feat(core): support st-global in custom property declaration

tzachbon:
## Tasks and info

- [x] `@property` supports `st-global`
- fix: `@st-global-custom-property` is not re-declaring `@property` 
    -  ~~remove warning~~ - we pass on this since it is going to be removed next major
    - [x] set global flag no matter the order of rules
- deprecate `@st-global-custom-property `
    - [x] codemod `@st-global-custom-property` --> `@property` list with no body
    - [x] deprecate warn `@st-global-custom-property`
- [x] validation: `@property` **with body** is valid [according to spec](https://drafts.css-houdini.org/css-properties-values-api/#at-property-rule) - "@property rules require a syntax and inherits descriptor; if either are missing, the entire rule is invalid and must be ignored. The initial-value descriptor is optional only if the syntax is the universal syntax definition, otherwise the descriptor is required; if it’s missing, the entire rule is invalid and must be ignored."
- [x] optimization: remove `@property` without body
- [x] update documentation https://github.com/wixplosives/stylable.io/pull/29

### Addiatinal cleanups 
- [x] separate each of the codemods tests to its own spec file.


packages/core/test/css-vars.spec.ts
```diff
@@ -58,6 +68,39 @@ describe('css custom-properties (vars)', () => {
...
+                    stGlobal(--b): red;
+
+                    color: var(stGlobal(--c), red);
+                }
+            `,
+                { from: 'path/to/style.css' }
+            );
+
+            expect(diagnostics.reports.length, 'no reports').to.eql(0);
+            expect(cssVars).to.eql({
```
idoros, line 85: what about the case where you use the same var both in `stGlobal` and without it?
it might be better just to prevent transformation, but not add anything to the metadata when simply using it like in this case.

barak007: What makes this argument stronger that this is a variable usage and not definition.

packages/core/src/stylable-processor.ts
```diff
@@ -345,10 +345,13 @@ export class StylableProcessor {
 
                     break;
                 case 'property':
-                    this.checkRedeclareSymbol(atRule.params, atRule);
                     this.addCSSVarDefinition(atRule);
                     break;
                 case 'st-global-custom-property': {
+                    deprecated(
```
tzachbon, line 351: @idoros does it this right way to deprecate it?

packages/core/test/stylable-transformer/at-property.spec.ts
```diff
@@ -98,4 +99,150 @@ describe('@property support', () => {
             },
         ]);
     });
+
+    describe('validation', () => {
+        it('should remove at property when used without a body', () => {
```
idoros, line 104: this is a transformation check - belongs outside of validation

packages/core/test/stylable-transformer/at-property.spec.ts
```diff
@@ -98,4 +99,150 @@ describe('@property support', () => {
...
+                        
+                        @property --y;
+                        `,
+                    },
+                },
+            };
+
+            const result = expectWarningsFromTransform(config, []);
+
+            const declarations = result.meta.outputAst!.nodes;
```
idoros, line 125: declaration is confusing. while a declaration might found at the top level, these are definitly at-rules or at least nodes

packages/core/src/index.ts
```diff
@@ -200,7 +201,7 @@ export { DiagnosticsMode, EmitDiagnosticsContext, emitDiagnostics } from './repo
 export { visitMetaCSSDependenciesBFS } from './visit-meta-css-dependencies';
 export { murmurhash3_32_gc } from './murmurhash';
 export { TimedCacheOptions, timedCache } from './timed-cache';
-
+export { atPropertyValidationWarnings } from './validate-at-property';
```
idoros, line 204: don't export internals - import from '@stylable/core/dist/...' when using in tests

packages/core/src/validate-at-property.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (!atRule.nodes?.length) {
+        return {
+            valid: true,
+            remove: true,
+        };
+    }
+
+    for (const node of atRule.nodes) {
+        if (node.type !== 'decl') {
```
idoros, line 32: why not warn on unknown content? we should check if other declaration/rules might break validity

packages/core/src/stylable-processor.ts
```diff
@@ -346,48 +352,35 @@ export class StylableProcessor {
...
                     this.addCSSVarDefinition(atRule);
+
+                    const { valid, message, remove } = validateAtProperty(atRule);
+
+                    if (!valid && message) {
+                        this.diagnostics.warn(atRule, message, { word: atRule.params });
+                    }
+
+                    if (remove) {
+                        toRemove.push(atRule);
```
idoros, line 367: move to the end of the transformer flow, we don't want to add more changes in the processor

packages/core/src/index.ts
```diff
@@ -30,6 +30,7 @@ export {
     processorWarnings,
     validateScopingSelector,
     parsePseudoImport,
+    parseStGlobalCustomProperty,
```
idoros, line 33: I know this is used outside the core, but I would prefer it being duplicated in the codemode and not adding another export for a deprecated feature (the duplicate code in the codemod would be kept long after the core would delete this code)

packages/cli/test/cli-codemod.spec.ts
```diff
@@ -83,4 +83,71 @@ describe('Stylable Cli Code Mods', () => {
             expect(stdout).to.match(new RegExp(processorWarnings.MULTIPLE_FROM_IN_IMPORT()));
         });
     });
+
+    describe('st-global-custom-property-to-at-property', () => {
```
idoros, line 87: maybe break into multiple codemod specs (one file per codemod)

packages/language-service/test/lib/diagnostics.spec.ts
```diff
@@ -158,7 +158,10 @@ describe('diagnostics', () => {
                 filePath
             );
 
-            expect(diagnostics).to.eql([]);
+            expect(diagnostics).to.have.length(1);
+            expect(diagnostics[0]).to.deep.include({
+                message: processorWarnings.DEPRECATED_ST_GLOBAL_CUSTOM_PROPERTY(),
+            });
```
idoros, line 164: I can see that the deprecation diagnostic is now emited, but the test doesn't make sense even before this change. It would be better to remove it imo.

idoros (changes requested):
Generally looks good. I think the deprecated syntax tests would have been better grouped together `describe("deprecated @st-global-custom-property")`.

I left comments about the 2 new exports from `core`, both are easy to change, but I think we should really talk about how we enable internal cross repository code share (for cases that are really required).

packages/core/test/css-vars.spec.ts
```diff
@@ -879,37 +933,70 @@ describe('css custom-properties (vars)', () => {
...
             expectWarningsFromTransform(config, [
+                {
+                    message: processorWarnings.DEPRECATED_ST_GLOBAL_CUSTOM_PROPERTY(),
+                    file: '/entry.st.css',
+                    severity: 'info',
+                },
                 {
                     message: processorWarnings.ILLEGAL_GLOBAL_CSS_VAR('illegalVar'),
                     file: '/entry.st.css',
+                    skipLocationCheck: true,
```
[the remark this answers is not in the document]
tzachbon: The expectation of location is acting weird when there are more than 2 diagnostics on the same node.

packages/core/src/validate-at-property.ts
idoros: The check for nested at-rules/rules is good, but I think the most helpful warning here is for unknown declaration. In the case someone would mistakenly types one of the intended declaration wrong (e.g. `initialvalue`)

idoros (approved):

barak007 (approved):
Great work.

PR #2014, opened 2021-08-16
refactor: rename keyframes "stGlobal" function to "st-global"

tzachbon:
We were rethinking the naming of our st function (`stGlobal`, `stArray`, `stMap`) and we would like to stick with the native CSS naming conventions ([CSS Functional Notation
](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions)).

Syntax:
```css
selector {
  property: functional-notation( [argument]? [, argument]! );
}
```


barak007:
what with `stMap` and `stArray` ? maybe we need to align all of them

PR #2016, opened 2021-08-17
fix(runtime): fix issue where stylable 4 can make stylable 3 stylessheets disappear

barak007:
So the story is:
When Stylable 3 runtime does reconciliation of the style elements (css-in-js mode) it checks the `st-id` attribute in order to verify that the style is actually from stylable. 
This is a weak check, and it happens to be that Stylable 4 also uses the same `st-id` attribute and Stylable 3 thinks that it can delete it and does so when they are siblings in the DOM.

This is the case when it happens
```html
<style st-id="22" st-depth="0">...</style> <!--Stylable 3 node-->
<style st-id="o28iZy" st-depth="0" st-runtime="0">...</style> <!--Stylable 4 node as sibling-->
```

This PR fixes this issue by changing the attributes from `st-id` to `st_id` this should avoid the collision.


barak007:
@tomrav please release after the review.

tomrav (approved):

PR #2018, opened 2021-08-18
feat(core): add CSS function notation syntax for "stMap" and "stArray"

tzachbon:
- [x] - remove `processDeclarationUrls` and use `processDeclarationFunctions` 
- [x] - add `st-array`, `st-map`
- [x] - show deprecation warning
- [x] - update documentation https://github.com/wixplosives/stylable.io/pull/28


packages/core/src/stylable-processor.ts
```diff
@@ -1055,3 +1088,33 @@ export interface DeclStylableProps {
 export interface SDecl extends postcss.Declaration {
     stylable: DeclStylableProps;
 }
+
+export function processDeclarationFunctions(
+    decl: postcss.Declaration,
+    onFunction: (node: AnyValueNode) => void,
```
tomrav, line 1094: why is the type used here (`AnyValueNode`) coming from `css-selector-tokenizer`?

Also, maybe these two utils should go in `stylable-utils.ts`?

packages/core/test/functions.spec.ts
```diff
@@ -678,6 +678,98 @@ describe('Stylable functions (native, formatter and variable)', () => {
     });
 
     describe('diagnostics', () => {
+        describe('deprecation', () => {
+            it('should return a warning when used deprecated stMap', () => {
```
tomrav, line 682: `should return a warning when using deprecated "stMap" syntax`

packages/core/test/functions.spec.ts
```diff
@@ -678,6 +678,98 @@ describe('Stylable functions (native, formatter and variable)', () => {
...
+                                'st-map'
+                            ),
+                            file: '/style.st.css',
+                            severity: 'info'
+                        },
+                    ]
+                );
+            });
+
+            it('should return a warning when used deprecated stArray', () => {
```
tomrav, line 710: `should return a warning when using deprecated "st-array" syntax`

packages/core/test/functions.spec.ts
```diff
@@ -678,6 +678,98 @@ describe('Stylable functions (native, formatter and variable)', () => {
...
+                                'st-array'
+                            ),
+                            file: '/style.st.css',
+                            severity: 'info'
+                        },
+                    ]
+                );
+            });
+
+            it('should return a warning when used deprecated nested stArray or stMap', () => {
```
tomrav, line 737: `should return a warning when using a nested deprecated "stArray" or "stMap" syntax`

tomrav (approved):
Left a couple of questions/suggestions, looks good otherwise.

barak007 (approved):

PR #2019, opened 2021-08-18
feat: support info diagnostic type

tzachbon:
- [x] - An option to emit info diagnostics that will always “warn” (also when the `diagnosticsMode` is `strict`).
- [x] - Handle it in `report-diagnostic` inside `core`.
- [x] - Handle it in the CLI.
- [x] - Test Webpack e2e flow
- [ ] - ~~Test CLI e2e flow~~ (Cannot be stubbed at the moment, will be added once it merge to [this PR](https://github.com/wix/stylable/pull/2018))


packages/webpack-plugin/test/e2e/projects/emit-info-diagnostic/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+    context: __dirname,
+    devtool: 'source-map',
+    plugins: [
+        new StylableWebpackPlugin({
+            diagnosticsMode: 'strict',
+            stylableConfig: (config) => ({
+                ...config,
+                hooks: {
+                    postProcessor: (result) => {
+                        result.meta.diagnostics.info(result.meta.ast.root(), 'test info warning!');
```
barak007, line 16: Add some info why this is here

barak007 (commented):
I feel good with this. I have one concern about adding `reportInfo` on the context...but if it will be required we can work it out quickly.

packages/cli/test/cli.spec.ts
```diff
@@ -268,6 +268,25 @@ describe('Stylable Cli', function () {
             expect(stdout, 'stdout').to.match(/unknown var "xxx"/);
         });
 
+        it.skip('(diagnosticsMode) should not exit with error when using strict mode with only info diagnostics', () => {
+            // Todo: test info diagnostic when we have one.
```
idoros, line 272: it's fine to test with available diagnostics, but even then I would add an issue about aligning the options between webpack and CLI ( keeping a tech-debt to replace with a test that won't break once some info diagnostics is removed). Having the same hook `postProcess` hook or any other hook would allow to have the same test as webpack with a custom non temporary info diagnostics.

idoros (approved):

packages/cli/test/cli.spec.ts
tzachbon: I would love to have that feature, and we will probably add it once we know how to parse the stc config from the stylable config. (we can add it to the research)
but, I would never prefer a test that stubs info diagnostic than emitting real info diagnostic.
Until we have that use case, I will add a `todo` there that will be updated once we have permanent info diagnostic.

tomrav (approved):
looks good to me.

Issue #1216, opened 2020-06-17
Support alternative suffix ".stcss" for source stylesheets


tomrav:
We've chosen to use the `.st.css` file suffix because it offers a mostly better developer experience in tools without explicit Stylable support (e.g. GitHub, IDEs without our extension, and so on). 
We get decent-ish syntax highlighting for free and even some of the native language service capabilities *may* work sporadically.

However, it does complicate integrating Stylable into any boilerplate or application that is already configured to work with `.css` files. This is made even harder when the configuration is external to the user project (for example any project based off of `create-react-app` or `next.js` ).

If we were to support an additional suffix, say, `.stcss`, we would lose the advantages stated above, but would be much easier to integrate to a variety of use-cases where we would no longer conflict with existing configurations.

I am not sure if this should be configured as supporting either suffix or both. I can imagine a use-case where the dependency tree is filled with a mixture of both suffixes.

PR #2031, opened 2021-08-29
feat: handle runtime types in a dedicated package

tzachbon:
Today when using stylable with typescript, the user will have to declare the `*.st.css` types globally.
This has a lot of limitations. The biggest one is that we have no control over it.

For example, if we support another extension, all the users will have to add the same code for `*.stcss` (https://github.com/wix/stylable/issues/1216).

This PR suggests creating a dedicated package that will handle all the `runtime-types`, and the user will need to specify it inside the `types` array inside the `tsconfig.json`.

```jsonc
{
    "compilerOptions": {
        //...
        "types": ["@stylable/runtime-types"],           
        //...
    }
}
```

### DIFF
--- a/packages/create-stylable-app/template/ts-react-rollup/template.js
+++ b/packages/create-stylable-app/template/ts-react-rollup/template.js
@@ -1,7 +1,8 @@
 /** @type {import('create-stylable-app').TemplateDefinition} */
 module.exports = {
-    dependencies: ['@stylable/runtime', 'react', 'react-dom'],
+    dependencies: ['react', 'react-dom'],
     devDependencies: [
+        '@stylable/runtime-types',
         '@rollup/plugin-commonjs',
         '@rollup/plugin-html',
         '@rollup/plugin-image',
--- a/packages/create-stylable-app/template/ts-react-rollup/tsconfig.json
+++ b/packages/create-stylable-app/template/ts-react-rollup/tsconfig.json
@@ -49,7 +49,7 @@
     // "paths": {},                                 /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
     // "rootDirs": [],                              /* List of root folders whose combined content represents the structure of the project at runtime. */
     // "typeRoots": [],                             /* List of folders to include type definitions from. */
-    "types": [],                                    /* Type declaration files to be included in compilation. */
+    "types": ["@stylable/runtime-types"],                                    /* Type declaration files to be included in compilation. */
     // "allowSyntheticDefaultImports": true,        /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
     "esModuleInterop": true,                        /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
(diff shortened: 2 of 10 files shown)


packages/create-stylable-app/template/ts-react-rollup/template.js
```diff
@@ -1,7 +1,8 @@
 /** @type {import('create-stylable-app').TemplateDefinition} */
 module.exports = {
-    dependencies: ['@stylable/runtime', 'react', 'react-dom'],
```
AviVahl, line 3: why remove this?

packages/create-stylable-app/template/ts-react-rollup/tsconfig.json
```diff
@@ -49,7 +49,7 @@
     // "paths": {},                                 /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
     // "rootDirs": [],                              /* List of root folders whose combined content represents the structure of the project at runtime. */
     // "typeRoots": [],                             /* List of folders to include type definitions from. */
-    "types": [],                                    /* Type declaration files to be included in compilation. */
+    "types": ["@stylable/runtime-types"],                                    /* Type declaration files to be included in compilation. */
```
AviVahl, line 52: comment indentation was aligned on purpose, for readability

packages/create-stylable-app/template/ts-react-webpack/template.js
```diff
@@ -1,7 +1,8 @@
 /** @type {import('create-stylable-app').TemplateDefinition} */
 module.exports = {
-    dependencies: ['@stylable/runtime', 'react', 'react-dom'],
```
AviVahl, line 3: why remove runtime?

packages/create-stylable-app/template/ts-react-rollup/template.js
tzachbon: Why did we need it in the first place (not as dev dep)? 
It was used for types as far as it looks.

tzachbon:
add indentation in the `tsconfig`

packages/create-stylable-app/template/ts-react-webpack/template.js
tzachbon: [Same answer](https://github.com/wix/stylable/pull/2031/files#r698008707)

packages/runtime-types/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '*.st.css' {
```
AviVahl, line 1: having a third party for this definition feels a bit odd. especially since it affects `*.st.css`. we don't usually let third parties augment `*.xyz` for us.

most typescript projects out there don't use "types", as it's not enabled by default for `tsc --init`.
if such projects turn it on, they will also have to specify other implicitly loaded types (node; jest; webpack-env; etc), making integration into existing projects harder.

tzachbon: Alright, I agree that it would be best to have the types without changing the "types" config but keeping it like today has many downsides. The other solution is to create a definitely typed package called something like `@types/stylable-runtime` / `@types/stylable-stylesheet`

@AviVahl  @barak007 wdyt?

tzachbon:
We'll wait with that for now.

Issue #1813, opened 2021-05-20
Cli optimization options


barak007:
Allow to specify optimization from cli arguments

PR #2032, opened 2021-08-30
test: emit info diagnostic from `st.css`

tzachbon:


packages/cli/test/cli.spec.ts
```diff
@@ -268,12 +268,10 @@ describe('Stylable Cli', function () {
...
         });
 
-        it.skip('(diagnosticsMode) should not exit with error when using strict mode with only info diagnostics', () => {
-            // Todo: test info diagnostic when we have one.
-            // https://github.com/wix/stylable/pull/2018
+        it('(diagnosticsMode) should not exit with error when using strict mode with only info diagnostics', () => {
             populateDirectorySync(tempDir.path, {
                 'package.json': `{"name": "test", "version": "0.0.0"}`,
-                'style.st.css': `.root {} `,
+                'style.st.css': `:vars { colors: stArray(red, blue); } `,
```
tomrav, line 274:
```suggestion
                'style.st.css': `:vars { colors: stArray(red, blue); }`,
```

packages/cli/test/cli.spec.ts
```diff
@@ -285,6 +283,7 @@ describe('Stylable Cli', function () {
             expect(status).to.equal(0);
             expect(stdout, 'stdout').to.match(/\[Stylable Diagnostics\]/);
             expect(stdout, 'stdout').to.match(/style\.st\.css/);
+            expect(stdout, 'stdout').to.match(/\[info\]: "stArray" is deprecated, use "st-array"/);
```
tomrav, line 286: use diagnostic function?

packages/webpack-plugin/test/e2e/emit-info-diagnostic.spec.ts
```diff
@@ -26,7 +26,7 @@ describe(`(${project})`, () => {
         const warnings = projectRunner.getBuildWarningMessages();
 
         expect(warnings).to.have.length(1);
-        expect(warnings[0]).to.match(/test info warning!/);
+        expect(warnings[0]).to.match(/"stArray" is deprecated, use "st-array"/);
```
tomrav, line 29: use diagnostic function?

packages/webpack-plugin/test/e2e/projects/emit-info-diagnostic/webpack.config.js
```diff
@@ -9,17 +9,6 @@ module.exports = {
     plugins: [
         new StylableWebpackPlugin({
             diagnosticsMode: 'strict',
```
tomrav, line 11: this is here to show that it doesn't break in strict mode with info diagnostics?

tomrav (commented):
Left a couple of questions

packages/webpack-plugin/test/e2e/projects/emit-info-diagnostic/webpack.config.js
tzachbon: yep

barak007 (approved):
I know we sad that we want to check it on a real info diagnostics but there is a downside for that.
When we will get no more infos we will not have something to test the behavior.

It's not a blocker for these tests. but the approach of emitting custom info seems more resilient to me.

tzachbon:
> I know we sad that we want to check it on a real info diagnostics but there is a downside for that.
> When we will get no more infos we will not have something to test the behavior.
> 
> It's not a blocker for these tests. but the approach of emitting custom info seems more resilient to me.

@barak007 alright then, we can wait with this until we have permanent info diagnostic or a way to hook to the post-process of the CLI #1813

tzachbon:
> I know we sad that we want to check it on a real info diagnostics but there is a downside for that.
> When we will get no more infos we will not have something to test the behavior.
> 
> It's not a blocker for these tests. but the approach of emitting custom info seems more resilient to me.

@barak007 I decided to keep the post-process hook until we will have a permanent `info` diagnostic.

barak007 (approved):

Issue #1996, opened 2021-08-08
CLI configuration file


tomrav:
### Description
Stylable currently supports one build flow.
We want the ability to structure a program into smaller pieces and give the ability to share configuration between them.

When working with multi packages project that uses Stylable CLI, you will need to call `stc` from each package and specify the configuration inline. For example, `stc --outDir dist`.
The solution that we are looking for is that the user will configure his project in one place (regardless of whether it is one package or many).

### Current main issues

* Projects are running `stc` multiple times and not once.
* There is no watch mode that understands the dependency graph and builds them in the correct package order.

### Goals

* Parsing configuration from a file `js` or `json` for the example -  `stylable.config.js`. Then the user can configure his project easily and use presets or share code (it also can be typed).
* Give the option to hook into process life cycles.
* Using one configuration from the root for the whole multi-package project.\
We can have a field named `projects` which can be an array of package names or a record that the key represents a request that can be a package name or a matching pattern and the value can be preset, presets, partial configuration, or object with preset and configuration override.
```ts
export type Presets = Record<string, PartialBuildOptions>;

export type Projects =
    | Array<string | [string, ProjectEntryValues]>
    | Record<string, ProjectEntryValues>;

export interface MultipleProjectsConfig extends Partial<SingleProjectConfig> {
    presets?: Presets;
    projects: Projects;
    projectsOptions?: {
        resolveRequests?: ResolveRequests;
    };
}


export type ProjectEntryValue<T extends string> =
    | T
    | PartialBuildOptions
    | {
          preset?: T;
          presets?: Array<T>;
          options: PartialBuildOptions;
      };

export type ProjectEntryValues<T extends string> =
    | ProjectEntryValue<T>
    | Array<ProjectEntryValue<T>>;


```
> By default the packages must be under `workspaces` field in the root package json.

* As stated above, we will give the option to declare presets available in the project.
We could also provide common presets that will have the options configured in the documentation for the user to see and use. For example, like `stcss-lib`, `css-lib`, `css-in-js-lib`, `ssr`, etc.

* The user can provide his own resolve requests method.

Preset examples:
* Usages in projects
```json
{
    "presets": {
        "stcss": { "dts": true, "extension": "stcss" }
    },
    "projects": {
        "lib": "stcss"
    },
}
```
* Usage in projects with override
```json
{
    "presets": {
        "stcss": { "dts": true, "extension": "stcss" }
    },
    "projects": {
        "lib": {
        	"preset": "stcss",
[...]

PR #2049, opened 2021-09-12
feat(cli): support configuration file

tzachbon:
This PR is part one of making CLI compatible with workspaces #1996 . 
The first thing we want to have is a way to resolve the configuration file and update the build method signature.

Opitions resolution order:
* defaults
* configuration file
* cli args

> The next step is resolving multiple projects.


packages/cli/src/logger.ts
```diff
@@ -16,4 +16,4 @@ export function createLogger(prefix: string, shouldLog: boolean) {
     };
 }
 
-export type Log = (...args: [...string[], symbol | string]) => void;
+export type Log = (...args: [...any[]]) => void;
```
barak007, line 19: Why changed?

tzachbon: The type is incorrect, we are passing there object too.

packages/cli/src/resolve-options.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+export function getCliArguments(): Arguments<CliArguments> {
+    const defaults = createDefaultOptions();
+    return yargs
+        .usage('$0 [options]')
+        .option('rootDir', {
+            type: 'string',
+            description: 'root directory of project',
+            default: process.cwd(),
```
barak007, line 45: move to defaults?

tzachbon: You can not configure the `rootDir` from the config. That's why it can not be in the defaults of the configuration options.

PR #2050, opened 2021-09-13
fix(core): js mixin re-export arguments parsing

barak007:
Currently are parsing the arguments of a any mixin in the process step and we figure out the strategy via the import type
if it's `.css` we choose `named` strategy for the rest `args`.

This is wrong since JS mixins can be exported from a stylesheet and then we choose the wrong strategy for them (`named`).
This PR keeps the old parsing of options (hopefully for a shot term) but reparse the args right after we actually know what type of mixin it is.

This PR also fixes an issue that we did not resolve the mixin deeply to figure out the origin.

### DIFF
--- a/packages/core/src/stylable-mixins.ts
+++ b/packages/core/src/stylable-mixins.ts
@@ -406,7 +406,7 @@ function reParseMixinNamedArgs(
     const options =
         mix.mixin.valueNode?.type === 'function'
             ? strategies.named(mix.mixin.valueNode, (message, options) => {
-                  diagnostics.warn(rule, message, options);
+                  diagnostics.warn(mix.mixin.originDecl || rule, message, options);
               })
             : (mix.mixin.options as Record<string, string>) || {};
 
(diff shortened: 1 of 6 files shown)


packages/core/src/stylable-mixins.ts
```diff
@@ -101,7 +110,7 @@ export function appendMixin(
                 );
             }
         } else {
-            // TODO: error cannot resolve mixin
+            // TODO: error cannot resolve mixin (this might be covered by unknown symbol)
```
idoros, line 113:
```suggestion
            // TODO: error cannot resolve mixin - this should be a diagnostic covered by unknown symbol
```

idoros (commented):
Maybe it would be better to completely separate the previous mixin analysis that we want to remove from the processor from the apply mixin in the transformer. You could keep the deprecated `mixins` data on the SRule in the processor for now, and just run the mixin parser again in the transformer process without adding or changing much in the processor or in the mixin parser (except removing the diagnostics).

packages/core/src/stylable-mixins.ts
```diff
@@ -61,16 +62,24 @@ export function appendMixin(
 
     const local = meta.mappedSymbols[mix.mixin.type];
```
idoros, line 63: local is a bit misleading as alias is also referenced under mappedSymbols

Issue #2084, opened 2021-10-04
[4.7.0] type error when using @stylable/core with typescript


AviVahl:
### Actual Behavior

Type error when using `@stylable/core` (in a typescript project).

### Expected Behavior

No type errors. Avoid depending on `@types` packages in public API.

### Version

@stylable/core@4.7.0

### Relevant log output

```shell
$ tsc
node_modules/@stylable/core/dist/stylable-utils.d.ts:7:30 - error TS7016: Could not find a declaration file for module 'css-selector-tokenizer'. '/home/avi/projects/stylable-intelligence/node_modules/css-selector-tokenizer/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/css-selector-tokenizer` if it exists or add a new declaration (.d.ts) file containing `declare module 'css-selector-tokenizer';`

7 import { AnyValueNode } from 'css-selector-tokenizer';
                               ~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

PR #2085, opened 2021-10-04
refactor(core): move internal functionality to separated index

barak007:
separate the function and import it directly from dist

fix: #2084


packages/core/src/internal.ts
```diff
@@ -0,0 +1,2 @@
+export * from './index';
```
tzachbon, line 1: I'm not sure if it should export all of the core too.

packages/core/src/internal.ts
```diff
@@ -0,0 +1,2 @@
...
+export * from './internal/process-declaration-functions';
```
tzachbon, line 2: Let's keep the explicit exports
```suggestion
export { processDeclarationFunctions } from './internal/process-declaration-functions';
```

tzachbon (commented):
In general, it looks good.
I left a couple of questions

packages/core/src/internal.ts
idoros: I agree

barak007: me too

barak007: @idoros should we?

barak007: maybe we can skip the internal.ts and just import files directly from the internal folder

idoros: I think it unifies the internal api in a way that can be simply mapped to package json exports when ts 4.5 will be out

PR #2086, opened 2021-10-04
feat: support CLI multiple projects

tzachbon:
This PR is part 2 of CLI improvements https://github.com/wix/stylable/issues/1996

## Goals

- [x] support multiple projects build.
- [x] build by topological order by default.
- [x] build multiple outputs for a specific request.
- [x] options override for a specific request.
- [x] creation of multiple presets.
- [x] override options of a specific request with preset.
- [x] override a preset value for a specific request.
- [x] handle watch mode.
- [x] watch cache invalidation optimization.
- [x] *BUG FIX* - copied assets and generated manifest do not mark as generated.
- [x] *BUG FIX* - graceful error handling for build processes.
- [x] create full "e2e" test with fake `node_modules`.
- [x] add top level file processor cache to all entities.
- [x] remove manually added `.js` extension from resolved paths when #2135 merged.
- [x] adjust logs in watch mode.
- [x] Update the `Generator` name to `IndexGenerator`.
- [x] Remove `extension` option.
- [ ] open tech-debt issue to move all log from CLI to `messages.ts`.
- [x] add to documentation.

## Changes
* Add new dependency `@wixc3/resolve-directory-context` for the default projects resolution.
* Export type helpers from the index.
* A lot of types were added and changed.
* Created messages file that contains all the output messages that we are using across the CLI.
* `visitMetaCSSDependenciesBFS` supports js dependency callback
* Config - split into 3 main files
  - `resolve requests` - the default `requests` resolution method. It will take the project raw entities and will output the resolved project in the relevant order.
  - `resolve options` - includes options calculation methods.
  - `process projects` - has the process project main method that resolves multi-project config to raw project entities.
  - `projects config` - this is the main file related to config resolution that exports the `projectsConfig` method which is the main method for resolving the projects.
 * Build -
   - Split the `build` method into two parameters (build options [defined by the user], and build metadata).
   - Each build has an instance of `DirectoryProcessService` which works only on its relevant scope.
   - After the initial build, the scopes have a map that includes all of the files dependent by list (the opposite of dependency).
  - The builds share a map of all of the files meta-data that understand each files `src` path and `out` paths.
  - In the top-level context there is a DirectoriesHandlerService which handles watch and watch cache.
* Currently we managed the diagnostic pretty much in a straight-forward way which does not work in complex situations that the multiple projects and builds introduced. A diagnostic manager was created to solve this issue by managing all the diagnostic in a high-level manager to be shared between all processes.

## Testing
- `generate-index` and `build` specs were changed to match the new `build` method API.
- `config-options` spec tests the resolution of options.
- `config-presets` spec test the resolution of presets and multiple presets.
- `config-project` spec tests all of the cases that can occur from multiple projects (resolution order, multi-options resolution, etc...)
- `cli-watch` refactor to support tests for single build and multiple builds.
- Moved `cli-test-kit` to `@stylable/e2e-test-kit`.

## Test helpers
  - the watch CLI test-kit now supports delay after the steps are completed (it is used to make sure in specific tests that there is no extra logs / circular build after the steps are finished).


packages/core/src/visit-meta-css-dependencies.ts
```diff
@@ -5,7 +5,8 @@ import type { StylableResolver } from './stylable-resolver';
 export function visitMetaCSSDependenciesBFS(
     meta: StylableMeta,
     onMetaDependency: (meta: StylableMeta, imported: Imported, depth: number) => void,
-    resolver: StylableResolver
+    resolver: StylableResolver,
+    onJsDependency?: (resolvedPath: string, imported: Imported) => void
```
idoros, line 9: why split the "onDependency" into 2 callbacks? wouldn't all usages benefit from also registering js dependencies (with a small change to the signature)?

packages/core/src/visit-meta-css-dependencies.ts
```diff
@@ -19,10 +20,18 @@ export function visitMetaCSSDependenciesBFS(
...
             if (res?._kind === 'css' && !visited.has(res.meta.source)) {
                 visited.add(res.meta.source);
                 onMetaDependency(res.meta, imported, depth + 1);
                 q[depth + 1].push(...res.meta.getImportStatements());
+            } else if (res?._kind === 'js' && onJsDependency) {
+                const resolvedPath = resolver.resolvePath(imported.context, imported.request);
+
+                if (!visited.has(resolvedPath)) {
+                    visited.add(resolvedPath);
+                    onJsDependency(resolvedPath, imported);
```
idoros, line 33: is there a limitation/issue when updating an imported module from a JS import (_CSS<-JS<-JS_)?

packages/cli/src/types.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+export interface ProjectsConfigResult {
+    rootDir: string;
+    projects: STCConfig;
+}
+
+export interface BuildOptions {
+    /** Specify the extension of stylable files */
+    extension: string;
```
idoros, line 130: knowing that we want to move to `.stcss`, maybe we shouldn't add support for choosing the extension

packages/cli/src/types.ts
```diff
@@ -0,0 +1,190 @@
...
+    /** specify where to find source files */
+    srcDir: string;
+    /** specify where to build the target files */
+    outDir: string;
+    /** should the build need to output manifest file */
+    manifest?: string;
+    /** opt into build index file and specify the filepath for the generated index file */
+    indexFile?: string;
+    /** custom cli index generator class */
+    Generator?: typeof BaseGenerator;
```
idoros, line 140: `IndexGenerator` is more specific

packages/cli/README.md
```diff
@@ -101,6 +167,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+```
+
+> The full types specification for defining Projects
+```ts
+export type Projects =
+    | Array<string | [string, ProjectEntryValues]>
+    | Record<string, ProjectEntryValues>;
+
+// P = Preset name
+export type ProjectEntryValue<P extends string> =
```
idoros, line 237:
```suggestion
export type ProjectEntryValue<PRESET extends string> =
```

packages/cli/README.md
```diff
@@ -101,6 +167,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+    | P
+    | PartialBuildOptions
+    | {
+          preset?: P;
+          presets?: Array<P>;
+          options: PartialBuildOptions;
+      };
+
+export type ProjectEntryValues<P extends string> =
```
idoros, line 246: maybe switch `ProjectEntryValues` to be above `ProjectEntryValue` as it is used first and then uses the other.

packages/core/src/visit-meta-css-dependencies.ts
tzachbon: This method is used externally, and we did not want to introduce breaking changes in this PR.
I do agree that this PR should be unified to a single `onDependency` method.

packages/core/src/visit-meta-css-dependencies.ts
tzachbon: Yep, we do not go deep inside the JS file, we might introduce this in feature.

packages/cli/test/cli-watch.spec.ts
```diff
@@ -1,380 +1,1074 @@
...
+                    },
+                ],
+            });
+
+            const files = loadDirSync(tempDir.path);
+            expect(files['packages/project-a/dist/style.css']).to.include('color:blue');
+            expect(
+                Object.keys(files).some((file) => file.includes(tempDir.path)),
+                'build dependency from the wrong scope (build from scope "b" inside scope "a")'
+            ).to.be.false;
```
barak007, line 610: to.eql(false)

packages/cli/test/cli-watch.spec.ts
```diff
@@ -1,380 +1,1074 @@
...
+                            1,
+                            sep + join('packages', 'project-a')
+                        ),
+                    },
+                ],
+            });
+
+            const files = loadDirSync(tempDir.path);
+            expect(files['packages/project-a/dist/style.css']).to.include('color:blue');
+            expect(
```
barak007, line 607: `validateNoExternalPathCreatedInTestDirectory`

packages/cli/test/cli-watch.spec.ts
```diff
@@ -1,380 +1,1074 @@
...
+                        ),
+                    },
+                    {
+                        msg: processMessages.FINISHED_PROCESSING(
+                            1,
+                            `[0] ${sep}` + join('packages', 'project-a')
+                        ),
+                        action() {
+                            return {
+                                sleep: 500,
```
barak007, line 866: add comment here

packages/cli/test/cli-watch.spec.ts
```diff
@@ -1,380 +1,1074 @@
...
-                },
-                {
-                    msg: messages.FINISHED_PROCESSING,
-                },
-            ],
+                    {
+                        msg: processMessages.FINISHED_PROCESSING(1),
+                    },
+                    {
+                        msg: processMessages.FINISHED_PROCESSING(1),
```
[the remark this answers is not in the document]
tzachbon: For the two processes, delete, and then creation.

barak007 (approved):
Great work

barak007 (approved):

packages/cli/README.md
```diff
@@ -54,6 +58,67 @@ After installing `@stylable/cli`, the `stc` command will be available, running `
 
 `*` - For the `useNamespaceReference` flag to function properly, the `source` folder must be published in addition to the output `target` code
 
+### Configuration file
+
+`stc` configuration file should be defined in the `stylable.config.js` under the name `stcConfig`.
```
tomrav, line 63:
```suggestion
The `stc` configuration should be located in the `stylable.config.js` file under the property name `stcConfig`.
```

packages/cli/README.md
```diff
@@ -54,6 +58,67 @@ After installing `@stylable/cli`, the `stc` command will be available, running `
...
+We also provide a helper method and types to give you a better configuration experience.
```
tomrav, line 64:
```suggestion
The CLI provides a helper method and type definitions to provide a better configuration experience.
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
 $ stc --srcDir="./src" --outDir="./dist"
 ```
 
+## Multiple Projects
+
+Projects allow you to share `stc` configuration and manage your Stylable projects in one place with predictable and controlled build order with caching optimizations.
```
tomrav, line 171:
```suggestion
Projects allow sharing `stc` configurations and management of Stylable projects in one location. They provides a controllable and predictable build order with caching optimizations.
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+        useNamespaceReference: true, 
+    },
+    projects: ['packages/*']
+});
+
+```
+
+### Options
+
+Like for [single project](#configuration-file) `options` is the top-level `BuildOptions` and is the default options for each project.
```
tomrav, line 203:
```suggestion
Similar to a [single project](#configuration-file), `options` is the top-level `BuildOptions` and is the default options for each project.
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+
+### Projects
+
+**Projects** is a generic term that refers to the set of requests with defined single or multiple `BuildOptions`.\
```
tomrav, line 207:
```suggestion
**Projects** is a generic term that refers to a set of path requests that define single or multiple `BuildOptions`.\
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+This set of requests is being processed and then evaluated as a map of `projectRoot` (directory path) to a set of `BuildOptions`.
+
+By default, the request is a path to a package, and in order to make the correct topological sort, the dependency needs to be specified in each package `package.json`
+
+As said, the value of the request can be resolved to a single `BuildOptions`, but it can also be multiple builds.
```
tomrav, line 212:
```suggestion
As mentioned above, the value of a request can be resolved to a single or multiple `BuildOptions`.
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+          presets?: Array<PRESET>;
+          options: PartialBuildOptions;
+      };
+```
+
+
+
+### Presets
+
+To reuse `BuildOptions`, you can define them under a name inside the `presets` property and use them inside the project entry value.
```
tomrav, line 254:
```suggestion
To reuse `BuildOptions`, define them using a name under the `presets` property and use them as the project entry value.
```

packages/cli/README.md
```diff
@@ -101,6 +166,119 @@ To transform your project stylesheets to target JavaScript modules containing th
...
+    projects: {
+        'packages/*': ['firstPreset', 'secondPreset']
+    }
+
+};
+```
+
+### Projects Options
+
+These options let you control the projects resolution process.
```
tomrav, line 272:
```suggestion
These options control the projects resolution process.
```

tomrav (commented):
left a couple of comments for the readme

PR #2087, opened 2021-10-05
fix(cli): module formats default resolutaion

tzachbon:
Fixes config file module formats.

### DIFF
--- a/packages/cli/src/resolve-options.ts
+++ b/packages/cli/src/resolve-options.ts
@@ -193,19 +193,14 @@ export function resolveCliOptions(
 ): PartialConfigOptions {
     const rootDir = argv.rootDir;
     const outDir = argv.outDir ?? defaults.outDir;
-    const moduleFormats = [
-        ...new Set([
-            ...(defaults.moduleFormats || []),
-            ...getModuleFormats({ esm: argv.esm, cjs: argv.cjs }),
-        ]),
-    ];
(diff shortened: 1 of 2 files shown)


barak007 (approved):

packages/cli/test/config.spec.ts
```diff
@@ -21,17 +21,21 @@ describe('Stylable Cli Config', function () {
...
+                    })
                 `,
             });
 
             runCliSync(['--rootDir', tempDir.path]);
 
             const dirContent = loadDirSync(tempDir.path);
             expect(Object.keys(dirContent)).to.eql([
-                'dist/style.st.css.js',
+                'dist/style.st.css.mjs',
```
barak007, line 38: might want to check that there is no `cjs` file

tzachbon: It checks for equality so these are the only files

packages/cli/src/resolve-options.ts
```diff
@@ -58,12 +58,12 @@ export function getCliArguments(): Arguments<CliArguments> {
...
             type: 'boolean',
             description: 'output esm module (.mjs)',
-            defaultDescription: 'false',
+            defaultDescription: String(defaults.esm),
         })
         .option('cjs', {
             type: 'boolean',
             description: 'output commonjs module (.js)',
-            defaultDescription: 'true',
+            defaultDescription: String(defaults.esm),
```
barak007, line 66: fix me

PR #2125, opened 2021-11-01
refactor(core): initial feature split

idoros:
This PR starts the refactor into features and help us understand the refactor process for the rest of the features. 

- [x] chore: `featureContext` interface to pass down `meta`, `diagnostics` and in the future other build data like `presetEnv` and `mode`
- [x] break: removed `meta.simpleSelectors` - most likely not used anywhere

### `st-symbol` feature

top level feature that manages stylesheet symbols (although top level, this feature will have type knowledge of higher level feature symbols in order to keep the type system simple)

- [x] extract from processor
- [x] extract from transformer
- deprecate `meta.mappedSymbols`
    - [x] add getter deprecation
    - [x] remove any internal usage 
    - [x] alternative public API is required
        - `meta.mappedSymbols` -> `meta.getSymbol(name)` & `meta.getAllSymbols()`

 
### `css-class` feature

manage stylesheet's css-classes

- [x] extract from processor
- [x] extract from transformer
- deprecate `meta.classes`
    - [x] add getter deprecation
    - [x] remove any internal usage 
    - [x] alternative public API is required
        - `meta.classes` -> `meta.getClass(name)` & `meta.getAllClasses()`

### `css-type` feature

manage stylesheet's native and component type selectors

- [x] extract from processor
- [x] extract from transformer
- deprecate `meta.elements`
    - [x] add getter deprecation
    - [x] remove any internal usage 
    - [x] alternative public API is required
        - `meta.elements` -> `meta.getTypeElement(name)` & `meta.getAllTypeElements()`


packages/core/src/helpers/plugable-record.ts
```diff
@@ -0,0 +1,26 @@
...
+  get<K extends Key>(map: PlugableRecord, key: K): Val<K> | undefined {
+    return map[key];
+  },
+  getAssure<K extends Key>(map: PlugableRecord, key: K): Val<K> {
+    if (!map[key]) {
+      throw new Error(`key ${key.description} is missing on map`);
+    }
+    return map[key];
+  },
+  getUnsafeAssure<K extends Key>(map: PlugableRecord, key: K): Val<K> {
```
tzachbon, line 23:
```suggestion
  getUnsafeAssure<K extends Key>(map: PlugableRecord, key: K): Val<K> | undefined {
```
?

idoros: thats the `unsafe` part of the method, `get` returns `Val<K> | undefined` and `getAssure` returns `Val<K>` or throws

tzachbon: I understand what it does, but `getUnsafeAssure` return type will always be `Val<K>` right now.

idoros: it is used by [places](https://github.com/wix/stylable/pull/2125/files#diff-1936f22036b3d48f2edc918dc1d252051ca2155259914cbe7f69412bda43cf09R54) that know for a fact that the value is set and allows them to ignore undefined check for type and runtime value

idoros: I changed the name to `getUnsafe`, which might be a bit more clear, what do you think?

packages/core/src/features/css-class.ts
```diff
@@ -0,0 +1,198 @@
...
+            // ToDo: refactor out to transformer validation phase
+            validateRuleStateDefinition(context.rule, meta, resolver, originMeta.diagnostics);
+        }
+        namespaceClass(meta, symbol, node, originMeta);
+    },
+});
+
+// API
+
+export function getClass({ data }: StylableMeta, name: string): ClassSymbol | undefined {
```
tomrav, line 81: unify with `getSymbols` API

tomrav:
Looks awesome, left some notes to go over.

packages/core/test/features/css-class.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+                                    -st-global: ".z.zz";
+                                }
+                            `,
+                        },
+                    },
+                });
+    
+                testInlineExpects(result);
+            });
+            it(`should replace imported type`, () => {
```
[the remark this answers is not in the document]
idoros: moving and adding a test for classes that checks `-st-global` transformation of imported classes

packages/core/src/stylable-transformer.ts
```diff
@@ -646,7 +601,7 @@ export class StylableTransformer {
                 });
             }
         } else if (node.type === `nesting`) {
-            const origin = originMeta.mappedSymbols[originMeta.root] as ClassSymbol;
+            const origin = STSymbol.getSymbol(originMeta, originMeta.root) as ClassSymbol;
```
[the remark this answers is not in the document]
idoros: This is correct to pick up the symbol from the general symbol table as the nesting is used by mixin roots that can be an imported type-element (that will resolve into a class eventually).

I added a comment that tries to explain it.

packages/core/src/features/css-class.ts
```diff
@@ -0,0 +1,198 @@
...
+    return classSymbol;
+}
+
+export function namespaceClass(
+    meta: StylableMeta,
+    symbol: StylableSymbol,
+    node: FunctionalSelector,
+    originMeta: StylableMeta
+) {
+    if (valueMapping.global in symbol) {
```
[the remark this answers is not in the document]
idoros: there is a check for value just a few lines into the if block, but it is more similar to the previous behavior to have the actual value check in the top condition. moved it up.

packages/core/src/features/css-class.ts
```diff
@@ -0,0 +1,198 @@
...
+            meta.classes[name] = classSymbol;
+        });
+    }
+    return classSymbol;
+}
+
+export function namespaceClass(
+    meta: StylableMeta,
+    symbol: StylableSymbol,
+    node: FunctionalSelector,
```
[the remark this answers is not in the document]
idoros: this is called for to transform 3 types: `class`, imported root `type-element` and `pseudo_element`, non are "functional" and the namespace class will get rid of the parentheses for `-st-global` in the process by converting them into `selector` types if they are anyway.

Changed the argument type to `SelectorNode` and the convert to `convertToSelector` instead of `flattenFunctionalSelector`.

packages/core/src/features/css-class.ts
idoros: unified features symbol api

- `getClass`, `getType`, `getPart`, `getSymbol` -> `get`
- `getSymbols` => `getAll`

I kept `addSymbol`, `addClass`, `addType` (and the non existing `addPart`) separate for now as they have different signatures and are easy to change later if we decide to do so, because they are only used internally in the core process.

packages/core-test-kit/src/diagnostics.ts
```diff
@@ -51,49 +58,106 @@ export function findTestLocations(css: string) {
...
+            }
+        });
 
-    expect(res.diagnostics.reports.length, 'diagnostics reports match').to.equal(warnings.length);
+        expect(res.diagnostics.reports.length, 'diagnostics reports match').to.equal(
+            warnings.length
+        );
+    }
+}
+export const expectAnalyzeDiagnostics = expectWarnings;
```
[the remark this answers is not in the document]
idoros: refactor:
- `expectWarningsFromTransform` -> `expectTransformDiagnostics`
- `expectWarnings` -> `expectAnalyzeDiagnostics`

packages/core/src/stylable-meta.ts
```diff
@@ -27,28 +33,31 @@ export class StylableMeta {
...
-    public mappedSymbols: Record<string, StylableSymbol>;
+    public classes: Record<string, ClassSymbol> = {};
+    public elements: Record<string, ElementSymbol> = {};
+    public mappedSymbols: Record<string, StylableSymbol> = {};
     public mappedKeyframes: Record<string, KeyframesSymbol>;
     public customSelectors: Record<string, string>;
     public urls: string[];
     public parent?: StylableMeta;
-    public transformDiagnostics: Diagnostics | null;
+    public analyzeDiagnostics: Diagnostics | null = null;
```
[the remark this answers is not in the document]
idoros: removed `meta.analyzeDiagnostics` and keeped `meta.diagnostics` as the analyze diagnostics except while meta is in the transform process.

We should refactor the diagnostics API on the `meta` to improve simplicity in the future

packages/core/src/features/st-part.ts
```diff
@@ -0,0 +1,123 @@
...
+                            return false;
+                        });
+                    }
+                }
+            }
+        );
+    }
+
+    const resolvedElements: Record<string, Array<CSSResolve<ClassSymbol | ElementSymbol>>> = {};
+    for (const k of Object.keys(CSSType.getAll(meta))) {
```
tzachbon, line 119: `k`?

packages/core/src/stylable-meta.ts
```diff
@@ -57,21 +64,50 @@ export class StylableMeta {
...
         this.customSelectors = {};
         this.urls = [];
         this.scopes = [];
-        this.simpleSelectors = {};
         setFieldForDeprecation(this, `mixins`, { objectType: `stylableMeta` });
         this.mixins = [];
         this.transformDiagnostics = null;
         this.transformedScopes = null;
     }
+    getSymbol(name: string) {
```
tzachbon, line 76: This is great APIs but you rarely use them

packages/core/src/stylable-processor.ts
```diff
@@ -376,7 +366,10 @@ export class StylableProcessor {
                             };
 
                             this.meta.cssVars[cssVar] = property;
-                            this.meta.mappedSymbols[cssVar] = property;
+                            STSymbol.addSymbol({
```
tzachbon, line 369: Shouldn't it be API on the meta?

packages/core/src/stylable-processor.ts
```diff
@@ -975,7 +864,7 @@ export class StylableProcessor {
         value: any
     ) {
         const name = selector.replace('.', '');
-        const typedRule = this.meta.mappedSymbols[name] as ClassSymbol | ElementSymbol;
+        const typedRule = STSymbol.get(this.meta, name) as ClassSymbol | ElementSymbol;
```
tzachbon, line 867: this.meta.getSymbol

packages/core/src/stylable-resolver.ts
```diff
@@ -313,11 +318,12 @@ export class StylableResolver {
             isElement: boolean
         ) => void
     ): Array<CSSResolve<ClassSymbol | ElementSymbol>> {
-        const bucket = isElement ? meta.elements : meta.classes;
+        const feature = isElement ? CSSType : CSSClass;
```
tzachbon, line 321: meta.getClass or meta.getTypeElement would be much clearer

packages/core/src/stylable-transformer.ts
```diff
@@ -98,18 +96,7 @@ export const transformerWarnings = {
...
-    CANNOT_EXTEND_UNKNOWN_SYMBOL(name: string) {
-        return `cannot extend unknown symbol "${name}"`;
-    },
-    CANNOT_EXTEND_JS() {
-        return 'JS import is not extendable';
-    },
-    UNKNOWN_IMPORT_ALIAS(name: string) {
-        return `cannot use alias for unknown import "${name}"`;
-    },
+    ...STPart.diagnostics,
```
tzachbon, line 99: In the processor you used the feature's diagnostics object and here you spread it, it should be the same all around

tzachbon:
I love those changes. I left some comments that relate to the differences in APIs usages

tzachbon (commented):
I love those changes. I left some comments that relate to the differences in APIs usages.

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
tomrav, line 1: Why this?

idoros: mistake

packages/core/src/features/st-part.ts
idoros: it was a direct copy from the transformer, it should have changed to `name`

packages/core/src/stylable-meta.ts
idoros: Not sure if rarely, they also used all around the monorep.

These are public APIs, they are tested (each in it's own feature spec), and are replacements for the deprecated `meta.mappedSymbols`, `meta.classes`, and `meta.elements` that are used out of Stylbale

packages/core/src/stylable-processor.ts
idoros: `addSymbol` is not public and is only available as an internal feature API, we currently only allow getting info from the meta, no editing API

packages/core/src/stylable-resolver.ts
idoros: Used to be that way, but after talking to @tomrav, we went on unifying the features symbol interface to make it easier in places like this. a more specific API is offered by the public API: `meta.getClass` and `meta.getTypeElement`

idoros: Sorry, misunderstood the comment.

Internally in the core I have access to the features themselves, so I don't bother going through the meta public API. this is done for both easier debugging and also that we might want to narrow the public API in the future (make it serializable, remove ASTs or other internal data).

packages/core/src/stylable-processor.ts
idoros: same as with the following comment on the `meta.getClass or meta.getTypeElement` to replace direct feature access. In the core I went directly to features without passing through the public APIs

packages/core/src/stylable-transformer.ts
idoros: fixed

packages/core/src/stylable-resolver.ts
idoros: accepted your initial change

barak007 (approved):

tomrav (approved):

PR #2128, opened 2021-11-01
refactor(core): extend resolver cache entity

tzachbon:
Today we use the resolver cache and the value of its entities is very limited.
This PR extends the current cache entity value.

#### Note:
I found a minor issue in the rollup test that used `.js` when did not need.

### DIFF
--- a/packages/core/src/stylable-resolver.ts
+++ b/packages/core/src/stylable-resolver.ts
@@ -31,24 +31,19 @@ export interface InvalidCachedModule {
     error: unknown;
 }
 
-export type CachedStylableMeta =
-    | InvalidCachedModule
-    | {
-          resolvedPath: string;
-          kind: 'css';
-          value: StylableMeta;
-      };
+export interface CachedStylableMeta {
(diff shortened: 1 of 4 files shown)


packages/core/src/stylable-resolver.ts
```diff
@@ -24,8 +24,24 @@ export type JsModule = {
...
+export interface CachedStylableMeta extends BaseCachedModule<StylableMeta> {
+    kind: 'css';
+}
+
+export interface CachedJsModule extends BaseCachedModule<JsModule> {
+    kind: 'js';
+}
+
+export type CachedModuleEntity = CachedStylableMeta | CachedJsModule;
+export type CachedModule = CachedModuleEntity['value'];
```
tzachbon, line 43: Not sure about this type (`CachedModule`), it's exported from the `index` and it is created not to break anyone who uses it.
I would really want to remove it from here and from the `index`.

@barak007

barak007 (approved):

PR #2140, opened 2021-11-04
infra: limit the number of jobs in the test runner

tzachbon:
Lately, we get timeouts now and then from "heavy" tests from machines with limited resources, this PR should fix it.

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,4 +1,7 @@
+const os = require('os');
+
 module.exports = {
     colors: true,
     'enable-source-maps': true,
+    jobs: Math.min(4, os.cpus().length - 1),
 };
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -3,5 +3,5 @@ const os = require('os');
 module.exports = {
     colors: true,
     'enable-source-maps': true,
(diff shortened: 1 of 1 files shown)


.mocharc.js
```diff
@@ -1,4 +1,7 @@
+const os = require('os');
+
 module.exports = {
     colors: true,
     'enable-source-maps': true,
+    jobs: Math.min(2, os.cpus().length - 1),
```
AviVahl, line 6: could a slow/virtual machine return `os.cpus().length === 1`?
which will in turn cause `jobs: 0`

also, this didn't solve it on @barak007 machine

tzachbon: It would be the same result as today (https://github.com/mochajs/mocha/blob/34643e4c0821aeb8d6977c1942bc106c9363789a/lib/nodejs/buffered-worker-pool.js#L50)

barak007: on my machine that failures where timeouts because overload (the meet was open)...This is the 6th time this PR does not fail.

AviVahl:
not sure this is the right solution at all.

PR #2146, opened 2021-11-07
infra: use npm@8 to install dependencies

AviVahl:
- comes with Node 16 LTS
- it now supports `"workspaces"` in root `package.json`
- release3.yml left using yarn, as that branch didn't migrate
- use Node 16 for publishing
- test on Node 16 as well


barak007:
Npn lock file is huge in comparison to yarn...

AviVahl:
> Npn lock file is huge in comparison to yarn...

Yes, they use json and represent a lot more information in the lock file (actual instances of versions and the resulting tree). A bit less painful with lerna being gone, along with its aging version tree.

AviVahl:
currently fails because of typescript-eslint's new release not being compatible with ts@4.4.4

PR #2151, opened 2021-11-08
refactor!?:  rewrite project runner startup process

barak007:
CI testing

* Separate the coupling between `watch` option to the temp directory isolation
* Remove all sideEffects from the constructor of ProjectRunner (`describe` time) and move them to mocha `before` time
  - Create temp directory
  - Load webpack config
* Better cleanup of temp directory
* Built-in - #2143 (sorry not based on same commit)
* Simplify startup with `prepareTestDirectory` extracted from mocha specific setup
* Use the real path of input directory
* Remove parallel test limit
* disable more cache response options for server 
* improve test speed 
  - no need to wait for network idle
  - one browser server 
 

- [x] should the browser server require be in mocharc ?


idoros:
This branch still fail on my machine. The only thing that made tests non flaky until now was to reduce the amount of parallel jobs Mocha is running (to 4 in my case).

packages/e2e-test-kit/src/project-runner.ts
```diff
@@ -153,25 +108,28 @@ export class ProjectRunner {
...
-        let done = false;
-        return new Promise<void>((res) => {
-            this.compiler?.hooks.afterDone.tap('waitForRecompile', () => {
-                if (done) {
-                    return;
-                }
-                done = true;
+        return new Promise<void>((res, rej) => {
+            if (!this.compiler) {
+                return rej();
```
AviVahl, line 126: add new Error and tell the reason

AviVahl:
`throwOnBuildError` used to default to `true`. now it defaults to `false`. Is there anything else less strict now?

packages/e2e-test-kit/src/project-runner.ts
AviVahl, line 126: new Error saying there reason?

tzachbon:
> This branch still fail on my machine. The only thing that made tests non flaky until now was to reduce the amount of parallel jobs Mocha is running (to 4 in my case).

I guess the middle ground here is env variable.

idoros:
> I guess the middle ground here is env variable.

Barak added some improvements since I wrote the comment. Rollup test went from 15s to 5s on my machine, running all the tests went from ~190s down to 70s, at most I get 4-5 chromium processes, and all test are passing without flakiness in the few times I run them. 

Amazing work!

packages/e2e-test-kit/src/project-runner.ts
```diff
@@ -69,109 +52,84 @@ export class ProjectRunner {
...
+        this.prepareTestDirectory();
+        return this.options.watchMode ? this.watch() : this.bundle();
     }
     public async bundle() {
         this.log('Bundle Start');
-        const webpackConfig = this.getWebpackConfig();
+        const webpackConfig = this.loadWebpackConfig();
         const compiler = webpack(webpackConfig);
         this.compiler = compiler;
-        // compiler.run = compiler.run.bind(compiler);
```
barak007, line 122: remove line

packages/e2e-test-kit/src/browser-server.ts
```diff
@@ -0,0 +1,14 @@
+import { BrowserServer, chromium } from 'playwright-chromium';
+
+let server: BrowserServer;
```
barak007, line 3: move to this

.mocharc.js
```diff
@@ -1,7 +1,5 @@
-const os = require('os');
-
 module.exports = {
     colors: true,
     'enable-source-maps': true,
-    jobs: Math.min(2, os.cpus().length - 1),
+    require: [require.resolve('@stylable/e2e-test-kit/dist/browser-server.js')],
```
barak007, line 4: Two options:
1. Move to package.json to not run in all tests. also add to the webpack-plugin 
2. look at process.cwd and enable when running from root or webpack-plugin

tzachbon: 1

barak007: I choose 2 because it will give better experience when working on single package.

PR #2159, opened 2021-11-14
fix(jest): correct export to match Jest API expectation

tomrav:
While adding the configuration options in #2097, I made a mistake and created an incompatible API. This PR restores functionality to our Jest integration.

This also highlights the fact that we do not have real e2e tests for this integration.


barak007 (approved):
LGTM.
Just make sure it is actually working because our test is not e2e.
This package is a good candidate to move from stylable repo. after that we can use the real jest in tests

PR #2168, opened 2021-11-16
fix(core): restrict functional nested selectors analyze

idoros:
We have a regression that can potentially pick up `custom pseudo states` with parameter which matches an `element` for a value with a first capital letter or a `class` for a value that starts with a dot (e.g. `:custom(Comp)` or `:custom(.part)`).

That might cause unintended symbol conflict that reports a `error/warning` diagnostics and even cause transpilation errors.

The fix:

- skip analyze of nested AST of unknown functional pseudo classes (that contain selectors)
- add the other native pseudo-classes to the restricted list - this is a breaking change that disallows states with the names:  `host`, `host-context`, `nth-child`, `nth-last-child`, `nth-of-type`, and `nth-last-of-type`. 

```css
:not(Comp.part) {}

:custom(Xxx.yyy) {}
```

In the example, `Comp` type and `part` class symbols are collected since `:not` is listed in `knownPseudoClassesWithNestedSelectors`, but `Xxx` type and the `yyy` class are not.

### DIFF
--- a/packages/core/src/native-reserved-lists.ts
+++ b/packages/core/src/native-reserved-lists.ts
@@ -51,9 +51,9 @@ export const nativePseudoClasses = [
 ];
 
 /**
- * list names of pseudo classes that cannot be override by custom states
- * // [breaking] ToDo: add names of general pseudo classes that are not specific to elements
- * // like: root, only-child, nth-of-type, nth-last-child, nth-last-of-type, only-of-type...
+ * list names of functional pseudo classes
+ * - cannot be override by custom states
+ * - might contain nested selectors
  */
 export const reservedPseudoClasses = [
     `not`,
@@ -62,6 +62,12 @@ export const reservedPseudoClasses = [
     `is`,
     `where`,
     `has`,
+    `host`,
+    `host-context`,
+    `nth-child`,
+    `nth-last-child`,
+    `nth-of-type`,
+    `nth-last-of-type`,
     // not native
     `global`,
     `local`,
--- a/packages/core/src/native-reserved-lists.ts
+++ b/packages/core/src/native-reserved-lists.ts
@@ -55,7 +55,7 @@ export const nativePseudoClasses = [
  * - cannot be overridden by custom states
  * - might contain nested selectors
  */
-export const reservedPseudoClasses = [
+export const reservedFunctionalPseudoClasses = [
     `not`,
     `any`,
     `matches`,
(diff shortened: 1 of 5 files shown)


barak007 (approved):

packages/core/src/native-reserved-lists.ts
```diff
@@ -51,9 +51,9 @@ export const nativePseudoClasses = [
 ];
 
 /**
- * list names of pseudo classes that cannot be override by custom states
- * // [breaking] ToDo: add names of general pseudo classes that are not specific to elements
- * // like: root, only-child, nth-of-type, nth-last-child, nth-last-of-type, only-of-type...
+ * list names of functional pseudo classes
+ * - cannot be override by custom states
```
tomrav, line 55:
```suggestion
 * - cannot be overridden by custom states
```

packages/core/test/stylable-processor.spec.ts
```diff
@@ -305,6 +306,31 @@ describe('Stylable postcss process', () => {
         expect(Object.keys(result.elements).length).to.eql(1);
     });
 
+    it('should not collect typed elements or classes in unknown functional selectors', () => {
+        const result = processSource(
+            `
+            :unknown(Unknown.unknown) {}
+            ${reservedPseudoClasses.map((name) =>
```
tomrav, line 313: casting this in a template string joins it with a comma (`,`), possibly a concern?

packages/core/test/stylable-processor.spec.ts
```diff
@@ -305,6 +306,31 @@ describe('Stylable postcss process', () => {
...
+        `,
+            { from: 'path/to/style.css' }
+        );
+
+        expect(Object.keys(result.elements)).to.not.include(`Unknown`);
+        expect(Object.keys(result.classes)).to.not.include(`unknown`);
+        expect(Object.keys(result.elements)).to.not.include(`El-global`);
+        expect(Object.keys(result.classes)).to.not.include(`cls-global`);
+
+        const allWithoutGlobal = reservedPseudoClasses.filter((name) => name !== `global`);
```
tomrav, line 327: not sure I understand the purpose of this set of tests

tomrav (approved):
Good fix, left a couple of questions.

packages/core/test/stylable-processor.spec.ts
idoros: It checks that any nested selectors within the known functional pseudo-class list are collected. With the exception of `:global` which Stylable doesn't pick selectors from.

packages/core/test/stylable-processor.spec.ts
idoros: good catch

tomrav (approved):

PR #2172, opened 2021-11-21
feat(core): stylable resolver cache filter

tzachbon:
We want to have the option to filter out specific cache entries instead of invalidating all the cache at once.

I also added more cache tests that were missing.

### DIFF
--- a/packages/core-test-kit/src/generate-test-util.ts
+++ b/packages/core-test-kit/src/generate-test-util.ts
@@ -11,9 +11,13 @@ import {
     StylableTransformer,
     createStylableFileProcessor,
     createDefaultResolver,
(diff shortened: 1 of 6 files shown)


barak007 (approved):

packages/core-test-kit/src/generate-test-util.ts
```diff
@@ -111,3 +115,22 @@ export function generateStylableRoot(config: Config) {
 export function generateStylableExports(config: Config) {
     return generateStylableResult(config).exports;
 }
+
+export function generateStyleableEnvironment(
```
tomrav, line 119: should be `generateStylableEnvironment`

PR #2173, opened 2021-11-22
feat(rollup-plugin):  support generate dts

hewel:
Add a options to support generate dts

### DIFF
--- a/packages/rollup-plugin/src/index.ts
+++ b/packages/rollup-plugin/src/index.ts
@@ -15,6 +15,7 @@ import {
 } from '@stylable/build-tools';
 import { resolveNamespace as resolveNamespaceNode } from '@stylable/node';
 import { StylableOptimizer } from '@stylable/optimizer';
+import { generateDTSContent } from '@stylable/module-utils';
 import decache from 'decache';
 import {
     emitAssets,
@@ -32,6 +33,7 @@ export interface StylableRollupPluginOptions {
     mode?: 'development' | 'production';
     diagnosticsMode?: DiagnosticsMode;
     resolveNamespace?: typeof resolveNamespaceNode;
+    dts?: boolean;
 }
 
 const requireModuleCache = new Set<string>();
@@ -56,6 +58,7 @@ export function stylableRollupPlugin({
     diagnosticsMode = 'strict',
     mode = getDefaultMode(),
     resolveNamespace = resolveNamespaceNode,
+    dts = false,
 }: StylableRollupPluginOptions = {}): Plugin {
     let stylable!: Stylable;
     let extracted!: Map<any, any>;
@@ -103,6 +106,11 @@ export function stylableRollupPlugin({
                 }
             }
             extracted.set(id, { css });
+
+            if (dts) {
+                const dtsContent = generateDTSContent({ meta, exports });
+                fs.writeFileSync(id.replace('.st.css', '.d.ts'), dtsContent);
+            }
 
             visitMetaCSSDependenciesBFS(
                 meta,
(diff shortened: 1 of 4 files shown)


barak007:
Hey, thanks for the PR. that's an important workflow feature. 
I'm interested to know what is your setup and what do you use Stylable for.

packages/rollup-plugin/src/index.ts
```diff
@@ -32,6 +33,7 @@ export interface StylableRollupPluginOptions {
     mode?: 'development' | 'production';
     diagnosticsMode?: DiagnosticsMode;
     resolveNamespace?: typeof resolveNamespaceNode;
+    dts?: boolean;
```
tzachbon, line 36: We would probably want to give the option to generate `dts` source map like in the CLI.

packages/rollup-plugin/src/index.ts
```diff
@@ -104,6 +107,11 @@ export function stylableRollupPlugin({
             }
             extracted.set(id, { css });
 
+            if (dts) {
+                const dtsContent = generateDTSContent({ meta, exports });
+                fs.writeFileSync(id.replace('.st.css', '.d.ts'), dtsContent);
+            }
+
```
tzachbon, line 114: It should be:
```js
`${id}.d.ts`
```

tzachbon (commented):
First of all, thanks for your contribution. 
This feature sounds nice and is something to consider.

You can see the implementation we made in the CLI ([here](https://github.com/wix/stylable/blob/baded1f5c7d59d6b6f8669656c72cc65f39ae313/packages/cli/src/build-single-file.ts#L149)). If we want to use the same logic, we need to move this to somewhere that it could be re-used.

Regardless, we would like to talk with you about your use case and setup.

packages/rollup-plugin/src/index.ts
hewel: Its that value id not end with `.st.css`

packages/rollup-plugin/src/index.ts
hewel: > First of all, thanks for your contribution. This feature sounds nice and is something to consider.
> 
> You can see the implementation we made in the CLI ([here](https://github.com/wix/stylable/blob/baded1f5c7d59d6b6f8669656c72cc65f39ae313/packages/cli/src/build-single-file.ts#L149)). If we want to use the same logic, we need to move this to somewhere that it could be re-used.
> 
> Regardless, we would like to talk with you about your use case and setup.

I 'm sure the CLI performs the same purpose. However, I only want to install the rollup plugin and not the CLI in order to produce DTS.
When I use TypeScript, the DTS comes in handy. I just want to use it right away.

hewel:
> Hey, thanks for the PR. that's an important workflow feature. I'm interested to know what is your setup and what do you use Stylable for.
Stylable is a fantastic tool that substantially improves the experience of writing CSS. It's what I utilize for my personal projects. It performs admirably. Simply said, I'd like to have a better Typescript experience.

packages/rollup-plugin/src/index.ts
tzachbon: Yes, it does. The id is the absolute path.
Therefore we need to extend it with `.d.ts` to tell typescript to look for the declaration file of this stylesheet.

tomrav:
hey @hewel sorry for the delay in responding to this, we were on vacation :)

I tried pushing a couple of changes to your branch, but it appears that I do not have the proper permissions to do so.
Let me know if you would rather giving me access, or that we relay over to you the needed changes.

Thank you for your contribution!

tomrav:
I'm closing this PR as we've now added this functionality to both our webpack and rollup intergartions.

See documentation about generating `.d.ts` files for Stylable [here](https://stylable.io/docs/getting-started/typescript-integration/), and documentation about our config file [here](https://github.com/wix/stylable/tree/master/packages/cli#configuration-file).

Hope this will be useful for you, we'd love to hear feedback on this feature.

PR #2178, opened 2021-11-22
use IStylableOptimizer for StylableWebpackPluginOptions

netanelgilad:
Currently the webpack plugin options expect `StylableOptimizer`, so a class or an extending class. But when a consumer has more than a single instance of `@stylable/optimizer` in their dependency tree in different versions, Typescript fails on the types of the different `StyleableOptmizer` classes not matching.

So I noticed you have an interface of IStylableOptmizer which could also be used for the webpack plugin options and prevents this problem as Typescript would accept different concrete classes of `StylableOptmizer` even if they come from different versions.


barak007:
will be fixed by #2181

PR #2181, opened 2021-11-24
fix: use IStylableOptimizer in webpack config

barak007:
This should improve type compatibility in edge case where you pin Stylable versions.

reason #2178

### DIFF
--- a/packages/optimizer/src/stylable-optimizer.ts
+++ b/packages/optimizer/src/stylable-optimizer.ts
@@ -82,56 +82,6 @@ export class StylableOptimizer implements IStylableOptimizer {
         );
     }
 
-    public rewriteSelector(
-        selector: string,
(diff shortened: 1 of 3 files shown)


packages/optimizer/src/stylable-optimizer.ts
```diff
@@ -207,6 +157,57 @@ export class StylableOptimizer implements IStylableOptimizer {
                   }
         );
     }
+
+    protected rewriteSelector(
```
idoros, line 161: anything changed in this function?

barak007: Nope

barak007: Made it protected

idoros (approved):

PR #2199, opened 2021-12-08
fix(webpack-plugin): fix support of browser field

barak007:
It turns out that the internal stylable resolver did not resolve stylable files **internal** requests according to the webpack target and the package.json browser field.

The test that checked this support was actually only tested the resolution of stylable requests from the JavaScript side. 

This PR fixes this issue but has one drawback:
JavaScript mixins and formatters are now also respect the browser field. This is not great because they run in node during the build but also not that bad because we can assume that no one is going to map them in the browser field.

### DIFF
--- a/packages/webpack-plugin/src/plugin.ts
+++ b/packages/webpack-plugin/src/plugin.ts
@@ -42,6 +42,7 @@ import type {
     EntryPoint,
     LoaderData,
     NormalModuleFactory,
+    ResolveOptionsWebpackOptions,
     StylableBuildMeta,
     StylableLoaderContext,
 } from './types';
@@ -255,7 +256,7 @@ export class StylableWebpackPlugin {
             return;
         }
 
-        const resolverOptions = {
+        const resolverOptions: ResolveOptionsWebpackOptions = {
             ...compiler.options.resolve,
             aliasFields:
                 compiler.options.resolve.byDependency?.esm?.aliasFields ||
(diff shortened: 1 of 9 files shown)


packages/webpack-plugin/src/plugin.ts
```diff
@@ -255,6 +255,13 @@ export class StylableWebpackPlugin {
             return;
         }
 
+        const resolverOptions = {
```
AviVahl, line 258: Explicit type

AviVahl (approved):

packages/webpack-plugin/test/e2e/browser-field.spec.ts
```diff
@@ -20,21 +20,27 @@ describe(`(${project})`, () => {
...
+    it('renders esm button with override', async () => {
         const { page } = await projectRunner.openInBrowser();
         const styleElements = await page.evaluate(browserFunctions.getStyleElementsMetadata);
+        const { backgroundColor, fontSize } = await page.$eval('button', (el: HTMLElement) => ({
+            fontSize: getComputedStyle(el).fontSize,
+            backgroundColor: getComputedStyle(el).backgroundColor,
+        }));
+        /*
+            Expecting the backgroundColor to be green 
+            makes sure that the ems style button is rendered
```
barak007, line 32:
```suggestion
            makes sure that the esm style button is rendered
```

PR #2201, opened 2021-12-09
feat(webpack-plugin): add ability to report DUPLICATE_MODULE_NAMESPACE as warning

barak007:
This PR add the ability to set DUPLICATE_MODULE_NAMESPACE to `warn` and not break the build

TODO: 
- [x] Add test

### DIFF
--- a/packages/webpack-plugin/src/plugin-utils.ts
+++ b/packages/webpack-plugin/src/plugin-utils.ts
@@ -364,8 +364,12 @@ export function getSortedModules(stylableModules: Map<NormalModule, BuildData | 
 
 export function reportNamespaceCollision(
     namespaceToFileMapping: Map<string, Set<NormalModule>>,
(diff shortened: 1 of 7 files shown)


packages/webpack-plugin/src/plugin-utils.ts
```diff
@@ -364,8 +364,12 @@ export function getSortedModules(stylableModules: Map<NormalModule, BuildData |
 
 export function reportNamespaceCollision(
     namespaceToFileMapping: Map<string, Set<NormalModule>>,
-    compilation: Compilation
+    compilation: Compilation,
+    ignore?: boolean | 'warn'
```
tomrav, line 368: I suggest simplifying this API to contain less types, probably a string enum normalized on entry

tomrav (approved):

PR #2203, opened 2021-12-14
feat: STSymbol setup to manage all symbols

idoros:
This PR enable the ability to manage all of Stylable symbols under the `st-symbol` feature.

## fix
- [x] report diagnostics on all conflicting symbol definitions
- [x] root cannot be overridden

## feat
- [x] namespace - each symbol is registered to a namespace and symbols only conflict if they are in the same namespace

## refactor
- [x] current features to manage their symbols under `st-Symbol`
- [x] move `redeclare symbol` diagnostics to a single place


packages/core/src/features/st-symbol.ts
```diff
@@ -56,22 +139,66 @@ export function addSymbol({
...
+    meta: StylableMeta;
+    symbol: filterSymbols<NamespaceToSymbolType['main']>;
+    localName?: string;
+}) {
+    const { byNS, byNSFlat, byType } = plugableRecord.getUnsafe(meta.data, dataKey);
+    const name = localName || symbol.name;
+    byNS.main.push({ name, symbol, safeRedeclare: false, ast: undefined });
+    byNSFlat.main[name] = symbol;
+    byType[symbol._kind][name] = symbol;
+    // ToDo: maybe override type record according to symbol._kind
```
tomrav, line 203: todo no longer relevant

tomrav (approved):

packages/core/src/features/st-symbol.ts
```diff
@@ -56,22 +139,66 @@ export function addSymbol({
...
+            }
+            flat[name].push(symbolDecl);
+        }
+        for (const name of collisions) {
+            for (const { safeRedeclare, ast } of flat[name]) {
+                if (!safeRedeclare && ast) {
+                    context.diagnostics.warn(ast, diagnostics.REDECLARE_SYMBOL(name), {
+                        word: name,
+                    });
+                }
```
barak007, line 178: what is there isn't ast don't we still want to do something?

barak007 (approved):

packages/core/src/features/st-symbol.ts
idoros: The only case that we have at the moment is the initial root that is registered for all stylesheets. I'm not sure what would we report on if a definition is not from a specific node in the AST. It would be odd to mark the entire AST as erroneous. Leaving it like it is for the moment (not different from the way it was before).

PR #2221, opened 2021-12-22
fix(optimizer): downgrade csso

barak007:
So it turns out `csso@5` which uses `css-tree@2` is kind of broken for our case that uses the `tpa-processor`. it removes whitespace where it (maybe) should not.

I also opened issue in csso repo:
https://github.com/css/csso/issues/447

We might not want to merge this and solve it from outside

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1868,19 +1868,6 @@
         "fastparse": "^1.1.2"
       }
     },
-    "node_modules/css-tree": {
-      "version": "2.0.4",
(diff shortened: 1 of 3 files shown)


barak007:
So after thinking about this issue a little more I think that we should not downgrade and instead allow our integrations to accept minifier from outside at first stage and remove the built-in minifier next major

barak007:
So it turns out we have a way to override the minify function from the outside so no work should be done here.

PR #2229, opened 2021-12-24
Add stylableNamespace to module buildInfo

dvdovenko:


packages/experimental-loader/src/stylable-runtime-loader.ts
```diff
@@ -27,8 +27,26 @@ const stylableRuntimeLoader: LoaderDefinition = function loader(content) {
...
     const [namespace, mapping] = evalStylableExtractModule(content);
 
+    addBuildInfo(this, namespace);
+
     return createRuntimeTargetCode(namespace, mapping);
 };
 
+function addBuildInfo(ctx: LoaderContext<any>, namespace: string) {
+    try {
+        if (ctx && ctx._module) {
```
barak007, line 37: This `if` will make it silently fail. either throw an error or remove the if

packages/experimental-loader/test/testkit/manifest-plugin.ts
```diff
@@ -0,0 +1,32 @@
...
+    return (module as NormalModule).resource !== undefined;
+};
+
+export class TestManifestPlugin {
+    apply(compiler: Compiler) {
+        compiler.hooks.compilation.tap(TestManifestPlugin.name, (compilation) => {
+            compilation.hooks.additionalAssets.tap(TestManifestPlugin.name, () => {
+                const data: Record<string, string> = {};
+                for (const module of compilation.modules) {
+                    if (isNormalModule(module)) {
```
barak007, line 14: change this `isNormalModule` to `isStylableModule` (or move resource.endsWith('.st.css') to the first `if`) and assume that you always have the `stylableNamespace` on the buildInfo. if you don't want to assume that you can throw instead

barak007 (changes requested):

packages/experimental-loader/src/stylable-transform-loader.ts
```diff
@@ -85,6 +85,18 @@ const stylableLoader: LoaderDefinition = function (content) {
         stylable.resolver
     );
 
+    try {
```
barak007, line 88: why did you added this here also? not sure about it was there any issue?

barak007: @dvdovenko ?

dvdovenko: @barak007 yes, I added it here because typescript types for `this._module` can be undefined / null

barak007 (approved):

packages/experimental-loader/test/testkit/manifest-plugin.ts
```diff
@@ -25,7 +25,7 @@ export class TestManifestPlugin {
                     }
                 }
 
-                compilation.assets['test-manifest.json'] = new sources.RawSource(
+                compilation.assets['test-manifest.json'] = new compiler.webpack.sources.RawSource(
```
dvdovenko, line 28: 👍

PR #2230, opened 2021-12-27
fix(language-service): stylable completions should be sorted above css completions

tomrav:
### Tasks

- [x] - set Stylable completions to a higher priority compared to the native CSS ones
- [x] - dedupe resulting completions giving Stylable priority
- [x] - create in-memroy testing mechanism for simpler testing
- [x] - add e2e test in `stylable-intelligence` - after merge and publish test can be merged downstream (wix/stylable-intelligence#756)

### DIFF
--- a/packages/language-service/src/lib/service.ts
+++ b/packages/language-service/src/lib/service.ts
@@ -298,7 +298,14 @@ export class StylableLanguageService {
 
         for (const stComp of stCompletions) {
             const lspCompletion = CompletionItem.create(stComp.label);
-            const ted: TextEdit = TextEdit.replace(
+
+            if (!groupedCompletions.has(lspCompletion.label)) {
+                groupedCompletions.set(lspCompletion.label, lspCompletion);
+            } else {
+                continue;
+            }
+
+            const textEdit: TextEdit = TextEdit.replace(
                 stComp.range
                     ? stComp.range
                     : new ProviderRange(
@@ -309,7 +316,7 @@ export class StylableLanguageService {
             );
             lspCompletion.insertTextFormat = 2;
             lspCompletion.detail = stComp.detail;
-            lspCompletion.textEdit = ted;
+            lspCompletion.textEdit = textEdit;
             // override sorting order to the top
             // todo: decide on actual sorting for all stylable completions
             lspCompletion.sortText = 'a';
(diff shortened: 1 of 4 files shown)


packages/language-service/src/lib/service.ts
```diff
@@ -295,41 +294,59 @@ export class StylableLanguageService {
...
-                                  Math.max(position.character - 1, 0)
-                              ),
-                              position
-                          ),
-                    typeof com.insertText === 'string' ? com.insertText : com.insertText.source
+        const groupedCompletions = new Map<string, CompletionItem>();
+
+        for (const stComp of stCompletions) {
+            const lspCompletion = CompletionItem.create(stComp.label);
+            const ted: TextEdit = TextEdit.replace(
```
tzachbon, line 301: What does `ted` means?

packages/language-service/src/lib/service.ts
```diff
@@ -295,41 +294,59 @@ export class StylableLanguageService {
...
-                return lspCompletion;
-            })
-            .concat(this.cssService.getCompletions(cleanDocument, position));
+            } else if (stComp.triggerSignature) {
+                lspCompletion.command = Command.create(
+                    'additional',
+                    'editor.action.triggerParameterHints'
+                );
+            }
+            if (!groupedCompletions.has(lspCompletion.label)) {
```
tzachbon, line 331: I guess this `if` could be moved to line 301.

packages/language-service/src/lib/service.ts
tomrav: "text edit"

packages/language-service/src/lib/service.ts
tomrav: moved up

packages/language-service/src/lib/service.ts
```diff
@@ -295,41 +294,63 @@ export class StylableLanguageService {
...
+
+        const cssCompletions = this.cssService.getCompletions(cleanDocument, position);
+
+        for (const cssComp of cssCompletions) {
+            const label = cssComp.label;
+
+            if (!groupedCompletions.has(label)) {
+                // CSS declaration property names have built in sorting
+                // at-rules, rules and declaration values do not
+                cssComp.sortText = cssComp.sortText || 'z';
```
barak007, line 348: Does this prevent compilation to be sorted by the label?

tomrav: No, we still beat it in the sort - this maintains their internal sorting.

barak007 (approved):

PR #2243, opened 2022-01-02
fix(create-stylable-app): remove css reset

AviVahl:
- the `all: unset` caused our own `font-family` styling to not get inherited.
- any chosen reset can be considered opinionated, so remove and let app developer pick his own instead.
- makes the `ts-react-webpack-lean` template leaner, with one less file to worry about.
- with these changes, the `ts-react-webpack` and the `ts-react-rollup` templates both look as originally intended.


tzachbon (approved):
Looks good

barak007 (approved):
I'm fine with that.

PR #2253, opened 2022-01-06
test(cli): change write file sync to use async write file

barak007:
try to avoid the cli flakiness  

- [x] revert the github actions before thinking about merge


tzachbon:
The re-build indexes fail on node 16 too.
https://github.com/wix/stylable/runs/4753387375?check_suite_focus=true

tzachbon:
https://github.com/livereload/livereload-site/blob/master/livereload.com/_articles/troubleshooting/os-x-fsevents-bug-may-prevent-monitoring-of-certain-folders.md

Interesting investigation, looks related.

tzachbon (approved):
Looks good IMO

PR #2255, opened 2022-01-10
Unify tool to test stylable core

idoros:
This PR adds a single test util `testStylableCore` to the `@stylable/core-test-kit` which provides a single function to test all cases. solves: #1987 .

## tasks
- [x] single entry input
- [x] inline test stylesheets
- [x] multiple file structure input 
- [x] return the processed stylesheets meta and exports
- [x] expose Stylable instance and filesystem
- [x] accept Stylable configuration
- [x] support default `requireModule` for JS
- [x] documentation
- [x] inline testing fixes and capabilities
  - add `@transform-remove` to test a node was removed in output
  - throw when transformation expectations are performed on removed nodes
  - fix parenthesis bug in `@rule` declaration expectation

### DIFF
--- a/packages/core-test-kit/src/test-stylable-core.ts
+++ b/packages/core-test-kit/src/test-stylable-core.ts
@@ -1,5 +1,5 @@
 import { testInlineExpects } from './inline-expectation';
-import { Stylable, StylableConfig, StylableExports, StylableMeta } from '@stylable/core';
+import { Stylable, StylableConfig, StylableResults } from '@stylable/core';
 import { createMemoryFs } from '@file-services/memory';
 import type { IDirectoryContents, IFileSystem } from '@file-services/types';
 import { isAbsolute } from 'path';
@@ -18,6 +18,13 @@ export type TestStylableConfig = Omit<
     resolveNamespace?: StylableConfig['resolveNamespace'];
 };
 
+/**
+ * The test function takes in a single '/entry.st.css' stylesheet string
+ * or a directory structure and then runs 2 phases
+ * 1. build stylesheets from entries in a configurable order
+ * 2. run inline test on all '.st.css' files found
+ * @param input single '/entry.st.css' string or file system structure
+ */
 export function testStylableCore(
     input: string | IDirectoryContents,
     options: Partial<TestOptions> = {}
@@ -37,8 +44,8 @@ export function testStylableCore(
     const allSheets = fs.findFilesSync(`/`, { filterFile: ({ path }) => path.endsWith(`.st.css`) });
     const entries = options.entries || allSheets;
 
-    // transform entries
-    const sheets: Record<string, { meta: StylableMeta; exports: StylableExports }> = {};
+    // transform entries - run build in requested order
+    const sheets: Record<string, StylableResults> = {};
     for (const path of entries) {
         if (!isAbsolute(path || '')) {
             throw new Error(testStylableCore.errors.absoluteEntry(path));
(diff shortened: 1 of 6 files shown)


packages/core-test-kit/src/test-stylable-core.ts
```diff
@@ -0,0 +1,66 @@
...
+        resolveNamespace: (ns) => ns,
+        ...(options.stylableConfig || {}),
+    });
+
+    // collect sheets
+    const allSheets = fs.findFilesSync(`/`, { filterFile: ({ path }) => path.endsWith(`.st.css`) });
+    const entries = options.entries || allSheets;
+
+    // transform entries
+    const sheets: Record<string, { meta: StylableMeta; exports: StylableExports }> = {};
```
tzachbon, line 41:
```suggestion
    const sheets: Record<string, StylableResults> = {};
```

packages/core-test-kit/src/test-stylable-core.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    // inline test
+    for (const path of allSheets) {
+        const meta = stylable.process(path);
+        if (!meta.outputAst) {
+            // ToDo: test
+            stylable.transform(meta);
+        }
+        testInlineExpects({ meta });
+    }
```
tzachbon, line 59: `entries` could equal `allSheets` so you are looping over there twice. 

```suggestion
    for (const path of entries) {
        if (!isAbsolute(path || '')) {
            throw new Error(testStylableCore.errors.absoluteEntry(path));
        }
        const meta = stylable.process(path);
        const { exports } = stylable.transform(meta);
        
        // inline 
        testInlineExpects({ meta });
                
        sheets[path] = { meta, exports };
    }
```

packages/core-test-kit/src/test-stylable-core.ts
```diff
@@ -0,0 +1,66 @@
...
+        createMemoryFs(typeof input === `string` ? { '/entry.st.css': input } : input);
+    const stylable = Stylable.create({
+        fileSystem: fs,
+        projectRoot: '/',
+        resolveNamespace: (ns) => ns,
+        ...(options.stylableConfig || {}),
+    });
+
+    // collect sheets
+    const allSheets = fs.findFilesSync(`/`, { filterFile: ({ path }) => path.endsWith(`.st.css`) });
```
tzachbon, line 37: The root will always be `/`?

tzachbon (commented):
Nice, left a couple of questions and suggestions.

packages/core-test-kit/src/test-stylable-core.ts
idoros: currently supporting in memory fs, if we ever want to run this on real fs then this will need to change

packages/core-test-kit/src/test-stylable-core.ts
idoros: `entries` is not equal to `allSheets` in case `options.entries` is provided and the filesystem contain `.st.css` files that are not referenced in `entries`.

The reason I went with separate phases was to run a process that is close to a real scenario - run analyze _(process)_ on the `entries` dependencies and then run build _(process/transform)_ on the `entries` in the specified order. After this process is complete run the inline tests in any order.

tzachbon: Cool, I would add a comment that explains that duplication since it is that clear unless you deep dive into the logic.

packages/core-test-kit/src/test-stylable-core.ts
tzachbon: So you can add it to the options object, optional `rootDir`option, and the default value would be `/`.

idoros: Will try

idoros: might extend in the future, but currently no core tests requires such API

PR #2262, opened 2022-01-12
fix(webpack-plugin): crash with webpack@5.66.0

AviVahl:
while the interface of GeneratorContext clearly states that getData is optional, webpack's new version uses it as it was required. this is one of the downsides of them using ts-in-js with less strictness.

fix is to provide it with an actual map and extract the data uri from there when it's available. kept regex-matching fallback for older versions.


tomrav:
someone is looking for a fight #2263 :fist_oncoming:

PR #2266, opened 2022-01-15
fix(core): complex selectors inside nested pseudo-classes

barak007:
Calling `scopeSelectorAst` is mutative to the input AST. The new selector parser refactor introduced a flaw that we missed - the original input nodes are still mutated, but any addition AST is not inserted into the original input.

This has caused a regression that misses the space added before most pseudo-elements. for example:
`:is(Btn::label)` should be `:is(.btn .label)` and not `:is(.btn.label)`

This PR changes the behavior to be fully mutative, so that any inner changes or additions will now change the original input.

In addition this PR also removes an unwanted space before nested selector: 
`:is(::part)` is now transformed to `:is(.part)` instead of `:is( .part)`

### DIFF
--- a/packages/core/src/stylable-transformer.ts
+++ b/packages/core/src/stylable-transformer.ts
@@ -489,10 +489,18 @@ export class StylableTransformer {
                 if (!symbol[valueMapping.root]) {
                     continue;
                 }
-
+                const isFirstInSelector =
+                    context.selectorAst[context.selectorIndex].nodes[0] === node;
                 const customSelector = meta.customSelectors[':--' + node.value];
                 if (customSelector) {
-                    this.handleCustomSelector(customSelector, meta, context, node.value, node);
+                    this.handleCustomSelector(
+                        customSelector,
+                        meta,
+                        context,
+                        node.value,
+                        node,
+                        isFirstInSelector
+                    );
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/stylable-transformer.ts
```diff
@@ -578,7 +586,10 @@ export class StylableTransformer {
                             : node.nodes
                     ) as Selector[];
                     const nestedContext = context.createNestedContext(innerSelectors);
-                    this.scopeSelectorAst(nestedContext);
+                    const transformedSelectors = this.scopeSelectorAst(nestedContext);
+                    for (let i = 0; i < transformedSelectors.length; i++) {
```
idoros, line 590: for of?

idoros (approved):

packages/core/src/stylable-transformer.ts
barak007: I need the index

tomrav: .entries :D

barak007: I like the plain loop here. I'm keeping it.

idoros: you do you

idoros, line 590: It bugs me that `scopeSelectorAst` is mutative to the inner selector nodes, but not to the selector itself. I would prefer it would mutate the entire selector. In the current state this patch will probably be required in other places as well.

Lets see if we can change the input selector that goes into `scopeSelectorAst`

PR #2276, opened 2022-01-18
fix: export `StVar` type

tzachbon:
Exporting `StVar` type since it's a public API

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -35,6 +35,7 @@ export {
     KeyFrameWithNode,
     ResolvedElement,
     StylableExports,
+    StVar,
     StylableResults,
     StylableTransformer,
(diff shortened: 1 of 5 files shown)


idoros (changes requested):
I don't think this should be exported out of index, just the transformer and the runtime/types

tzachbon:
> I don't think this should be exported out of index, just the transformer and the runtime/types

@idoros I do not agree with that statement.
Since `StylableExports` is something that the user Interacts with, he might want to declare a `stVar` variable and use its type.
It's not a Stylable internals type.

barak007 (approved):

barak007 (changes requested):
change to RuntimeStVar

idoros (approved):

Issue #2002, opened 2021-08-08
Unify Stylable API 


tomrav:
Avoid using global directives, add:
- @st-vars
- @st-custom-selector
- etc...

tzachbon:
# Research Template

### description
Namespacing our APIs is used for being safe from conflicts with exists CSS APIs or user's APIs.
Unified API can help a user when using the APIs (for example, when misspelling property and we suggest a fix). But, also to the user who reads the code, he can easily distinguish between regular CSS and stylable keywords.

### tags
#core #feature #language-service

### related issues
- #1928

### tasks overview
In this task, we will add our `st` prefix to each of our existing APIs and slowly deprecate the old APIs.

* Add the new `st` APIs, and add deprecation warning to each of the old APIs.
* Update the existing documentation to use the new API.
* Add language service support.
* Codemod.
* Major after that, we will remove the old APIs.

### blocked by (dependencies)
 - #1999
 - Deciding on how to deprecate the old APIs gradually.
 - #1997

tomrav:
Great breakdown, I like this alignment - I think it'll make our lives easier in the long term.

Concerning custom-selectors, I think we should prefix that as well because looking at the current draft I'm not sure we support all of its capabilities, and it might further change before officially making it into the spec.

PR #2278, opened 2022-01-19
refactor: replace pseudo import syntax to st-import in Stylable fixtures

tzachbon:
As part of the Unify Stylable API #2002 issue, we want to switch all the old pseudo import syntax to the new st-import syntax.


tzachbon:
Updating the test files was a very painful job. I will revisit this once we refactor some stuff in the language service.

PR #2301, opened 2022-02-01
chore: upgrade postcss

barak007:
* fixed parsing of custom properties


AviVahl (approved):
also upgraded by dependabot at #2304

PR #2313, opened 2022-02-06
feat(core): added origin filepath to processNamespace

### LINKED ISSUE #2286: Remove the requirement for the referenced file in `st-namespace-reference` to be present
### Proposal Description

Currently it depends on the namespace resolver.
Our basic node namespace resolve uses `findConfig` from the referenced file and that's breaking if the file does not exist.
We can avoid it if we will pass the origin file to the resolver and find the config from there.  

### Additional Context

_No response_

barak007:
fixes #2286 

- [x] add test somehow

### DIFF
--- a/packages/cli/test/cli.spec.ts
+++ b/packages/cli/test/cli.spec.ts
@@ -1,10 +1,10 @@
-import { join } from 'path';
+import { join, dirname, relative } from 'path';
 import { expect } from 'chai';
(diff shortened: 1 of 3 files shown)


AviVahl:
What's origin?

barak007:
origin is the path of the file in the dist before we used the namespace reference switch-up

PR #2315, opened 2022-02-07
refactor(module-utils): 10x speed-up for generate dts tests

barak007:
use `noLib:true` since we don't use anything from the outside world in the generate `d.ts` tests

### DIFF
--- a/packages/e2e-test-kit/src/dts-kit.ts
+++ b/packages/e2e-test-kit/src/dts-kit.ts
@@ -51,7 +51,7 @@ export class DTSKit {
                 target: ScriptTarget.ES2020,
                 strict: true,
                 types: [],
-                lib: ['lib.dom.d.ts', 'lib.es2020.d.ts'],
+                noLib: true,
             },
             rootNames: [filePath],
         });
--- a/packages/module-utils/test/generate-dts.spec.ts
+++ b/packages/module-utils/test/generate-dts.spec.ts
@@ -303,7 +303,7 @@ describe('Generate DTS', function () {
                 "Type 'string' is not assignable to type 'boolean | undefined'"
             );
             expect(diagnostics).to.include(
-                "Type 'true' is not assignable to type 'string | undefined'"
+                "Type 'boolean' is not assignable to type 'string | undefined'"
             );
         });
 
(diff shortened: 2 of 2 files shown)


AviVahl:
May want to validate there are no diagnostics when dom lib is included. Can use skipDefaultLibCheck instead to skip rechecking dom types themselves.

barak007:
The majority of the tests does not include anything from output world so they don't need any libs. We have one test with dom libs  enabled but now it's specified in the test.

barak007:
These tests used to take ~30 sec in CI....

AviVahl:
Why did the diagnostics change if the types aren't used? I think the Record type is not loaded or sethong like that. Is the test verifying there are no new diagnostics?

barak007:
> Why did the diagnostics change if the types aren't used? I think the Record type is not loaded or something like that. Is the test verifying there are no new diagnostics?

No...I will check. 
The change is because the infer of the record `{state1: true}` changed from `true` to `boolean` I'm not sure why and I'm not sure that it matters.

barak007:
No new errors....most of the tests are also expecting `''` as the output

AviVahl:
The types use the Record type: https://github.com/wix/stylable/blob/master/packages/runtime/src/types.ts#L26
I'd really expect that to show up as another diagnostic, and the test to catch that.
Instead of the `noLib`, use `noDefaultLibCheck` and specify the types always. wouldn't that be fast enough?

barak007:
`noDefaultLibCheck` helps a little. it cuts it in half.

barak007:
> The types use the Record type: https://github.com/wix/stylable/blob/master/packages/runtime/src/types.ts#L26 I'd really expect that to show up as another diagnostic, and the test to catch that. Instead of the `noLib`, use `noDefaultLibCheck` and specify the types always. wouldn't that be fast enough?

it does if you used it directly. that's interesting will do more checks

barak007:
I will go on the safe side and bring back the `lib.es2020.d.ts` with the `noDefaultLibCheck` it still 20x faster!. I had the wrong measurements. 10x @AviVahl

AviVahl:
just noticed that `NodeRenderer` also uses `Element`.
**all** .d.ts tests that use the stylesheet runtime types are supposed to have `dom` then.

barak007:
> Edit
> Hide
> Delete

the `d.ts` does not use these types. actually nothing uses these types from the renderer.

AviVahl (approved):

PR #2319, opened 2022-02-07
fix(cli): relative target file path url

tzachbon:
* Relative target file path url in the source-maps target.
* Adding namespace-reference comment at the button of the generated file.
* Use `relative` instead of replace


barak007 (approved):

packages/module-utils/test/sourcemap.spec.ts
```diff
@@ -65,9 +64,9 @@ describe('.d.ts source-maps', () => {
         });
 
         const dtsText = generateDTSContent(res);
-        const sourcemapText = generateDTSSourceMap(dtsText, res.meta, 'src');
+        const sourcemapText = generateDTSSourceMap(dtsText, res.meta, './src/entry.st.css');
```
barak007, line 67: remove `./` from test to match real world behavior.

packages/cli/src/build-single-file.ts
```diff
@@ -175,10 +175,18 @@ export function buildSingleFile({
         // .d.ts.map
         // if not explicitly defined, assumed true with "--dts" parent scope
         if (dtsSourceMap !== false) {
+            const relativeTargetFilePath = relative(
+                dirname(outSrcPath),
+                outputSources ? outSrcPath : filePath
```
barak007, line 180: we might want to fix the source-map to account the namespace reference line we add.
or move that line to the bottom.

packages/cli/src/build-single-file.ts
```diff
@@ -67,7 +67,7 @@ export function buildSingleFile({
     diagnosticsMode = 'loose',
     diagnosticsManager = new DiagnosticsManager({ log }),
 }: BuildFileOptions) {
-    const { basename, dirname, join, relative, resolve } = fs;
+    const { basename, dirname, join, relative, resolve, isAbsolute } = fs;
     const outSrcPath = join(fullOutDir, filePath.replace(fullSrcDir, ''));
```
barak007, line 71: use relative(fullSrcDir, filePath)

Issue #1972, opened 2021-08-05
`-st-global` does not export correctly 


tzachbon:
### Actual Behavior

When using `-st-global`, the declaration does not transform as expected, but it does transform in the javascript exports.

[Playground link](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28*%21%27.y%2520%280%2520%2520%2520%2520-st-global%253A%2520%255C%27.x%255C%27%253B0%29000%27%29%7Eselected%21*%27%29*%27%252Findex.st.css0%255Cn%25010*_)

### Expected Behavior

Transform in both CSS and javascript.

### Version

4.5.0

idoros:
The behavior this issue reports is opposite of the current observed behavior - currently the transformation works as expected, while the JS exports provide the namespaced class name instead of the `-st-global` selector. see [reproduction](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%280%21%27.y%2520%28%2520%252F*%2520replaced%2520with%2520%25602%2560%2520selector%2520value%25203%2520%2520%2520%25202%253A%2520%255C%27.x%255C%27%253B1%2911%252F****%2520js%2520export%2520expected%2520%2560y%253Ax%2560%2520***3%27%29%7Eselected%210%27%290%27%252Findex.st.css1%255Cn2-st-global3*%252F1%25013210_)

Fixing this should resolve the value exported for a simple class, but we need to handle problematic cases:
- composed classes: `.a.b` -> `a b`
- complex selector: `.a .b` ->
    - option 1: report `complex selectors is not supported in -st-global` and maybe export empty or `.b`
    - option 2: export `.b` and create a custom selector for `::part` -> `& .a .b`
- native states: `.a:hover` ->
    - option 1: report `pseudo-class in not supported in -st-global` and maybe export empty or `.a`
    - option 2: export `.a` and create a custom selector for `::part` -> `& .a:hover`

Issue #1906, opened 2021-07-06
:global selector ignores commas


idoros:
placing commas in `:global` selector doesn't produce any warnings and removes it during transform.

```css
:global(a, b) {}

/*transforms to*/

a b {}
```
[playground example](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28*%21%21global%257Ba%252C%2520b%257D%2520%28-%2520%2520%2520%2520-%29--%27%29%7Eselected%21*%27%29*%27%252Findex.st.css-%255Cn%2501-*_)

I think expected behavior should be to produce an error and don't transform the `:global` at all.

--- edit ---

`:global(a, b)` transforms the selector as is and omits the `,` resulting in an unintended selector transformation. The issue is that a merging-in the internal complex selector can be transformed in more then one way. For example `.a :global(.b, .c) .d` can be transformed to `.a .b, .c .d` or `.a .b .d, .a .c .d`, since there is no one correct answer, complex selectors should not be transformed and report a `complex selector is not supported in :global`.

tomrav:
Another option would be to create the required permutations for all cases `a, b {}`

idoros:
Thought about this, but something like:
```css
.x :global(a, b) .y {}
/*-->*/
.x a, b .y {}
```
This feels like a mistake. the initial parsing, done by a computer or a human, assumes a single selector, and the transformation break it into 2.

tomrav:
Maybe it should transpile it to the following?

```css
.x a .y,
.x b .y {}

```

idoros:
I think that is a bit confusing. splitting the selector like this is hard to explain and is not very predictable. 
Supporting a single complex selector is good enough in my opinion.

I would add a warning and an error in a next major.

btw, css-modules has the exact behavior (dropping the comma).

idoros:
related to #310

PR #2329, opened 2022-02-13
fix(core): css-class feature fixes 

idoros:
## fix
- [x] move CSS class resolve reports to transform diagnostics (mistakenly moved to analyze diagnostics)
- [x] correctly export `-st-global` classes to Javascript - #1972 
  - compound classes: `.a.b` -> `a b`
  - complex selector: `.a .b` -> `undefined` JS export
  - non class selector: `[attr]` -> `undefined` JS export
- [x] crash on empty `-st-global` - [playground](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28*%21%27.y%2520%28%25200%2520%2520%2520%2520-st-global%253A%2520%255C%27%255C%27%253B0%290%27%29%7Eselected%21*%27%29*%27%252Findex.st.css0%255Cn%25010*_)
- [x] report unsupported error for multi selector `-st-global` - [playground](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28*%21%27.a%2520%28%25200%2520%2520%2520%2520-st-global%253A%2520%255C%27.y%2520%252C%2520.x%255C%27%253B0%290%27%29%7Eselected%21*%27%29*%27%252Findex.st.css0%255Cn%25010*_)
- [x] fix missing diagnostic and problematic transformation for `:global()` with multiple selectors - #1906 

## refactor
- [x] move class JS exports code to feature (including css-class resolve)
- [x] move `:global` transform logic to the `st-global` feature

## test
- [x] refactor to use new `testStylableCore` test kit
- [x] add missing tests 
  - exports
  - `-st-extends`
  - `:global()`
  - import redeclare root diagnostic


packages/core/src/features/st-global.ts
```diff
@@ -1,17 +1,56 @@
...
+            context.diagnostics.info(rule, diagnostics.UNSUPPORTED_MULTI_SELECTOR_IN_GLOBAL(), {
+                word: stringifySelector(node.nodes),
+            });
+        }
+        return walkSelector.skipNested;
+    },
+    transformInit({ context }) {
+        context.meta.globals = {};
+    },
+    transformCleanup({ context: { meta } }) {
```
barak007, line 40: transformLastPass

packages/core/src/stylable-transformer.ts
```diff
@@ -175,14 +176,22 @@ export class StylableTransformer {
             resolver: this.resolver,
             evaluator: this.evaluator,
         };
+        const metaParts = this.getMetaParts(meta);
+        const cssClassResolve = CSSClass.hooks.transformResolve({
+            context: transformContext,
```
barak007, line 181: context once

packages/core/src/stylable-value-parsers.ts
```diff
@@ -83,13 +85,17 @@ export const SBTypesParsers = {
...
         const selector = parseSelectorWithCache(
             decl.value.replace(/^['"]/, '').replace(/['"]$/, ''),
             { clone: true }
         );
-        // ToDo: handle or warn on multiple selectors
+        if (!selector[0]) {
+            diagnostics.error(decl, CSSClass.diagnostics.EMPTY_ST_GLOBAL());
+            return;
+        } else if (selector.length > 1) {
+            diagnostics.error(decl, CSSClass.diagnostics.UNSUPPORTED_MULTI_SELECTORS_ST_GLOBAL());
```
barak007, line 97: should be info

packages/core/test/features/css-class.spec.ts
```diff
@@ -1,483 +1,718 @@
...
+    it(`should process css class selectors`, () => {
+        const { sheets } = testStylableCore(`
+            /* @rule(single) .entry__a */
+            .a {}
+
+            /* @rule(multi) .entry__b, .entry__c */
+            .b, .c {}
+
+            /* @rule(complex) .entry__d .entry__e*/
+            .d .e {}
```
barak007, line 46: add in media and st-scope

packages/core/test/features/css-class.spec.ts
```diff
@@ -1,483 +1,718 @@
...
         });
-        it(`should have root class symbol by default`, () => {
-            const { meta } = generateStylableResult({
-                entry: `/entry.st.css`,
-                files: {
-                    '/entry.st.css': {
-                        namespace: `entry`,
-                        content: ``,
+    });
+    it(`should add to general symbols`, () => {
```
barak007, line 107: unify with previous

packages/core/test/features/css-class.spec.ts
```diff
@@ -1,483 +1,718 @@
...
+        expect(CSSClass.get(meta, `a`), `symbol`).to.contain({
+            _kind: `class`,
+            name: 'a',
+        });
+
+        // JS exports
+        expect(exports.classes.a, `JS export`).to.eql(`entry__a`);
+    });
+    describe(`st-global`, () => {
+        it(`should override with :global() value`, () => {
```
barak007, line 309: rename

packages/core/test/features/css-class.spec.ts
```diff
@@ -1,483 +1,718 @@
...
+            });
+
+            // meta.globals
+            expect(meta.globals).to.eql({
+                r: true,
+                p: true,
+            });
+
+            // JS exports
+            expect(exports.classes.iRoot, `root alias JS export`).to.eql(undefined);
```
barak007, line 555: chack d.ts for alias root

barak007: hasOwnProperty?

idoros: `d.ts` usees exports to generate the type, so at the before this change the `d.ts` created an key for the aliased root, although the value is an empty string in reality, so this change will fix this issue as well - imported aliased root will not show as key under exported JS classes or in the `d.ts`

packages/core/src/stylable-value-parsers.ts
idoros: discussed this with Tom who thinks that both should be error for now - what do you think?

packages/core/test/stylable-transformer/exports.spec.ts
```diff
@@ -209,191 +31,6 @@ describe('Exports to js', () => {
...
-                        content: `
-                            .y{}
-                        `,
-                    },
-                },
-            });
-            expect(cssExports.classes.root).to.equal('entry__root index__y');
-        });
-
-        it('should handle multiple extends levels', () => {
```
barak007, line 322: maybe this also

packages/core/src/stylable-value-parsers.ts
idoros: changed to error with no transformation in case of multiple selectors. We can add support when `:is` is enabled in the future.

PR #2342, opened 2022-02-20
feat: run `stc` from integrations

tzachbon:
We want to give our users a better experience with the integrations and the `stc` tool.
Today, the user will probably have to run them in different processes to use them together.
This PR introduces a solution. If the integration finds a `stcConfig`, it will "merge" the processes programmatically, which will result in better diagnostics experience and consistent build order.

## Integrations (The CLI will become a dependency)

- [x] Support `webpack` integration.
- [x] Support `rollup` integration and add `2.70.0` of rollup as `peerDependency` of the package
- [x] Update README files.

## Feats
- [x] `loadStylableConfig` returns the path of the config too.
- [x] Create hooks API for the diagnostic manager. 
- [x] `reportDiagnostic` accepts `from` to the diagnostic report.
- [x] CLI supports "config" flag.
- [x] unified method to resolve `stcConfig`
- [x] `watchMode` flag for `stc` that does not trigger watch but tells `stc` to be prepared for watching.

## Fixes

- [x] Log skipped file in `stc` watcher.
- [x] `stc` watcher now knows about the first process's generated files.

## Refactor
- [x] move all `e2e-test-kit` file-system related helpers to a dedicated file.

## Test
- [x] fix dedup and sort diagnostics test in CLI.

PR #2343, opened 2022-02-21
feat(cli): use file services resolver in build stylable

tzachbon:
Has mentioned [here](https://github.com/wixplosives/file-services/issues/24#issuecomment-1014338289), this gives us a boost in terms of performance which we like in our CLI for big projects that are going to use our [integrations (e.g., webpack, rollup) with `stc`](https://github.com/wix/stylable/pull/2342).
The downside is that the resolver has missing features, but they are very advanced, and most modern tools still do not support them as well.

### DIFF
--- a/packages/cli/src/resolve-module.ts
+++ b/packages/cli/src/resolve-module.ts
@@ -1,0 +1,21 @@
+import type { IFileSystem } from '@file-services/types';
+import { createRequestResolver } from '@file-services/resolve';
+import type { StylableConfig } from '@stylable/core';
+
+export function createDefaultResolveModule(fs: IFileSystem): StylableConfig['resolveModule'] {
+    const moduleResolver = createRequestResolver({ fs });
+
+    return (context, request) => {
+        const { resolvedFile } = moduleResolver(context, request);
+
+        if (resolvedFile === false) {
+            throw new Error(
+                `Stylable CLI does not support browser field 'false' values. "${request}" resolved to 'false' from "${context}"`
+            );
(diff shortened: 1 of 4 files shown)


packages/cli/src/resolve-module.ts
```diff
@@ -0,0 +1,21 @@
+import type { IFileSystem } from '@file-services/types';
+import { createRequestResolver } from '@file-services/resolve';
+import type { StylableConfig } from '@stylable/core';
+
+export function createDefaultResolveModule(fs: IFileSystem): StylableConfig['resolveModule'] {
```
AviVahl, line 5: This is a much larger type than the resolver needs/uses. See it's options. There's another, smaller interface.

packages/cli/src/resolve-module.ts
```diff
@@ -0,0 +1,21 @@
...
+    const moduleResolver = createRequestResolver({ fs });
+
+    return (context, request) => {
+        const { resolvedFile } = moduleResolver(context, request);
+
+        if (resolvedFile === false) {
+            throw new Error(
+                `Stylable CLI does not support browser field 'false' values. "${request}" resolved to 'false' from "${context}"`
```
AviVahl, line 13: How is stylable core handling it?

AviVahl: nvmind https://github.com/wix/stylable/blob/master/packages/core/src/module-resolver.ts#L38

tomrav:
Closing for now (keeping branch).

PR #2380, opened 2022-03-08
infra: run tests for integrations and core packages separately

tzachbon:
A really simple test-runner tailored to our specific needs that uses mocha test runner behind the scenes.

### What we need:
* We want to run our integrations and core packages one after the other in CI.
* A way to run tests for selected packages from the workspace CWD.
* Remove unused launch configurations


barak007 (approved):

barak007:
let's talk with @AviVahl before merge.

PR #2393, opened 2022-03-15
feat(core): get `st-vars` computed values

tzachbon:
New public API to get the computed Stylable vars for a given style sheet.

### Internals
* Add the ability to require Stylable core package from the `core-test-kit`.
* Expose `flattenValue` from `CustomValueExtension` to indicate that the custom value can be flat.
* Move custom values resolution from each function evaluation to the resolver.

### DIFF
--- a/packages/core/src/features/st-var.ts
+++ b/packages/core/src/features/st-var.ts
@@ -3,8 +3,8 @@ import { deprecatedStFunctions } from '../custom-values';
 import { generalDiagnostics } from './diagnostics';
 import * as STSymbol from './st-symbol';
 import type { StylableMeta } from '../stylable-meta';
-import type { CSSResolve } from '../stylable-resolver';
-import type { EvalValueData, EvalValueResult } from '../functions';
+import { createSymbolResolverWithCache, CSSResolve } from '../stylable-resolver';
+import { EvalValueData, EvalValueResult, StylableEvaluator } from '../functions';
 import { isChildOfAtRule } from '../helpers/rule';
 import { walkSelector } from '../helpers/selector';
 import { stringifyFunction, getStringValue, strategies } from '../helpers/value';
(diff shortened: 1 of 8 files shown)


packages/core/src/index.ts
```diff
@@ -20,6 +20,7 @@ export type {
 } from './features';
 export { reservedKeyFrames } from './features/css-keyframes';
 export { scopeCSSVar } from './features/css-custom-property';
+export type { ComputedStVar, ComputedStVars } from './features/st-var';
```
idoros, line 23: remove record

packages/core/src/stylable.ts
```diff
@@ -101,6 +102,9 @@ export class Stylable {
 
         this.resolver = this.createResolver();
     }
+    public getComputedStVars(meta: StylableMeta) {
```
idoros, line 105: `Stylable.stVar.getComputed()`

packages/core/test/features/st-var.spec.ts
```diff
@@ -1255,4 +1259,145 @@ describe(`features/st-var`, () => {
         });
     });
     it.skip(`should provide valid var path introspection - value(var, ...path)`);
```
idoros, line 1261: remove skipped test that you replaced

idoros (changes requested):

idoros (approved):

packages/core/src/features/st-var.ts
```diff
@@ -112,6 +120,61 @@ export function get(meta: StylableMeta, name: string): VarSymbol | undefined {
     return STSymbol.get(meta, name, `var`);
 }
 
+// Stylable StVar Public APIs
+
+export class StylablePublicApi {
```
barak007, line 125: I don't like the Public in the name of this class. If it's exported it's public.

PR #2397, opened 2022-03-20
fix: cases with mixin moved compound selector

idoros:
This PR fixes 2 issues:

## 1) ambiguity of a mixed-in CSS class and root class 

The issue happens when a mixed class is compounded with `.root:state` (root class with `custom-pseudo-class`). The issue is that there is a bug in the process of mixin that loses the actual root symbol when a stylesheet class other then root is mixed:

**mixin.st.css:**
```
.root {
    -st-mixin: x;
}
.root:x.mix {
    color: green;
}
```
**entry.st.css:**
```
@st-import [mix] from "./mixin.st.css";

/* @rule[1] .entry__c.mixin__root.mixin--x */
.c {
    -st-mixin: mix;
}
```

valid selector: `.entry__c.mixin__root.mixin--x`
error selector: `.entry__c.mixin__root:x` // root symbol is "wrongly" replaced with mix class and `:x` custom-pseudo-state is unresolved

## 2) multiple compound selectors after mixin

When compound selectors are connected after mixin class, they need to move to the beginning of the selector with the mixin if it is moved:

**mixin.st.css:**
```
.root {
    -st-mixin: x;
}
.root.mix:mix-state {
    color: green;
}
```
**entry.st.css:**
```
@st-import [mix] from "./mixin.st.css";

/* @rule[1] .entry__c.mixin--mix-state.mixin__root */
.c {
    -st-mixin: mix;
}
```

valid selector: `.entry__c.mixin--mix-state.mixin__root`
error selector: `.entry__c.mixin__root:mix-state` // mix-state is not moved/transformed

### DIFF
--- a/packages/core/src/stylable-mixins.ts
+++ b/packages/core/src/stylable-mixins.ts
@@ -234,7 +234,8 @@ function createMixinRootFromCSSResolve(
         undefined,
         resolvedArgs,
         path.concat(symbolName + ' from ' + meta.source),
-        true
+        true,
+        resolvedClass.symbol.name
     );
 
     fixRelativeUrls(mixinRoot, mixinMeta.source, meta.source);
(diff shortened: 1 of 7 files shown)


packages/core/src/stylable-mixins.ts
```diff
@@ -338,21 +339,17 @@ function handleLocalClassMixin(
...
         path.concat(mix.mixin.type + ' from ' + meta.source),
-        true
+        true,
+        mix.ref.name
     );
     mergeRules(mixinRoot, rule);
 }
 
-function createInheritedMeta({ meta, symbol }: CSSResolve) {
+function createInheritedMeta({ meta }: CSSResolve) {
```
barak007, line 348: try remove

packages/core/src/stylable-transformer.ts
```diff
@@ -302,23 +303,27 @@ export class StylableTransformer {
...
+        rule?: postcss.Rule,
+        mixinClassName?: string
     ): { selector: string; elements: ResolvedElement[][]; targetSelectorAst: SelectorList } {
         const context = new ScopeContext(
             originMeta,
             this.resolver,
             parseSelectorWithCache(selector, { clone: true }),
             rule || postcss.rule({ selector })
         );
+        if (mixinClassName) {
```
barak007, line 324: move to new ScopeContext

PR #2402, opened 2022-03-22
fix(core): custom-value inside computed st-var

tzachbon:


idoros (commented):
looks good, but the computed input "type" field string might be problematic since it is only the local custom value identifier in a specific stylesheet. For example `stBorder` might be `border`, `stBorder` or `st-border`, depending on how it is imported, and when looking at a var that is a composite from imported vars it would be impossible to figure out what custom value it is actually made out of.

Maybe adding the module path and export or the actual evaluated custom-value would help?

packages/core/test/features/st-var.spec.ts
```diff
@@ -1284,7 +1284,42 @@ describe(`features/st-var`, () => {
             });
             expect(computedVars.c).to.containSubset({
                 value: ['red', 'gold'],
-                input: undefined,
+                input: {
+                    type: 'st-array',
+                    value: ['red', 'gold'],
```
idoros, line 1289: maybe `red` should be in the structure of `{ type: "string", value: "red" }`

packages/core/test/features/st-var.spec.ts
```diff
@@ -1284,7 +1284,42 @@ describe(`features/st-var`, () => {
...
+            :vars {
+                map: st-map(a st-map(b red));
+            }
+            `);
+
+            const { meta } = sheets['/entry.st.css'];
+            const computedVars = stylable.stVar.getComputed(meta);
+
+            expect(Object.keys(computedVars)).to.eql(['map']);
+            expect(computedVars.map).to.containSubset({
```
idoros, line 1306: add `flatValue` expectation

packages/core/test/features/st-var.spec.ts
```diff
@@ -1332,9 +1367,97 @@ describe(`features/st-var`, () => {
             expect(computedVars.border).to.containSubset({
                 value: '1px solid red',
                 input: {
-                    color: 'red',
-                    size: '1px',
-                    style: 'solid',
+                    type: 'stBorder',
+                    flatValue: '1px solid red',
```
idoros, line 1371: Add case where size is a map or array. In that case the flat value might not produce an expected result and should be omitted: [example](https://wixplosives.github.io/stylable-playground/#state=%28%27files%21%28I%21%27%2540Q-imporY%255BF%255D%2520from%2520R.%252FbWRJ%253Avars%2520Nb7F%257B1pxP43b27F%257BQ-array%257B1px62px%257DP%2944.x%2520NbK93b2SkK26sk6093%252F*%2520should%2520noYallow%2520flaYv8*%252F43b2K29%29%27%7E%252FbW.js%21%27conQ%2520ZMG6X_L%2520require%257BR%2540Qylable%252FcoreR9exports.FL%2520MG%257BNprocessArgs7H5X.argsH93%29EMO%255BD%255D5N33skE33QyleE33colorE3%29J3%29EgetOv86U5v8%255BU%255DJ3%29EflattenOZv87ZD__5N33delimiter7%2522%2520%2522E33parts7%255BD%255DE3%29J3%29%252C4%29%257D%27%29%7Eselected%21I%27%293%2520%2520%2520%25204%255Cn5%257DL%253E%2520N3return%25206%252C%25207%253A%25208alue9%257DJDsk6Qyle6colorE%252C43FQBWGCuQomV8H%257Bnode6cuQomTypesI%27%252FU.Q.cssJ%253B4K7v8%257BbL%2520%253DMcreateN%2843OV87%257BP6solid6red9QstR%255C%27UindexWorderXGStrategyYt%2520Z%28%2520_%2520%29kize%2501k_ZYXWURQPONMLKJIHGFED9876543_)

idoros (changes requested):

tzachbon:
Continued in this PR - https://github.com/wix/stylable/pull/2422

PR #2412, opened 2022-03-27
refactor(test-kit): remove `memory-fs` dependency

idoros:
This PR removes dependency on `memory-fs` and replace it with `@file-services/memory` where used.

In addition unused `@stylable/e2e-test-kit` code is removed (mem-fs/webpackTest/evelCssJsModule/createMemoryFileSystemWithFiles)


packages/module-utils/test/test-kit.ts
```diff
@@ -49,3 +50,15 @@ export function moduleFactoryTestKit(
         evalStylableModule,
     };
 }
+
+function addStylableRuntimeToMemFs(fs: IMemFileSystem) {
```
tzachbon, line 54: This method should not be coupled to `IMemFileSystem` if it only use `ensureDirectorySync` and `writeFileSync`.

idoros: it's a local function that is called from a single place, I prefer not to complicate the type

barak007 (approved):

PR #2415, opened 2022-03-28
webpack-plugin: package.json should declare "postcss" dep, because src/plugin.ts uses it directly

jakutis:


barak007:
10x for the report we somehow missed it for a year!

PR #2422, opened 2022-03-29
feat(core): flatten st-vars

tzachbon:
We want to give the option to flat the computed st-vars so they could be used for code modifications.

### DIFF
--- a/packages/core/src/custom-values.ts
+++ b/packages/core/src/custom-values.ts
@@ -3,6 +3,12 @@ import postcssValueParser from 'postcss-value-parser';
 import { getFormatterArgs, getNamedArgs, getStringValue } from './helpers/value';
 import type { ParsedValue } from './types';
 
+export class ValueError extends Error {
+    constructor(message: string, public fallbackValue: string) {
+        super(message);
+    }
+}
+
 export interface Box<Type extends string, Value> {
     type: Type;
     value: Value;
@@ -23,17 +29,19 @@ export function box<Type extends string, Value>(
 
 export function unbox<B extends Box<string, unknown>>(
     boxed: B | string,
-    customValues?: CustomTypes
+    unboxPrimitives = true,
+    customValues?: CustomTypes,
+    node?: ParsedValue
 ): any {
     if (typeof boxed === 'string') {
-        return boxed;
-    } else if (typeof boxed === 'object' && boxed?.type) {
+        return unboxPrimitives ? boxed : box('string', boxed);
+    } else if (typeof boxed === 'object' && boxed !== null) {
         const customValue = customValues?.[boxed.type];
         let value = boxed.value;
-        if (customValue?.flattenValue) {
-            value = customValue.getValue([], boxed, null, customValues!);
+        if (customValue?.flattenValue && node) {
+            value = customValue.getValue([], boxed, node, customValues!);
         }
-        return cloneDeepWith(value, (v) => unbox(v, customValues));
+        return cloneDeepWith(value, (v) => unbox(v, unboxPrimitives, customValues, node));
     }
 }
 
@@ -51,20 +59,21 @@ export interface CustomValueExtension<T> {
         valueAst: ParsedValue,
         customTypes: {
             [typeID: string]: CustomValueExtension<unknown>;
-        }
+        },
+        boxPrimitive?: boolean
     ): Box<string, T>;
     getValue(
         path: string[],
         value: Box<string, T>,
-        node: ParsedValue | null,
+        node: ParsedValue,
         customTypes: CustomTypes
     ): string;
 }
 
 function createStArrayCustomFunction() {
     return createCustomValue<BoxedValueArray, BoxedValueArray>({
-        processArgs: (node, customTypes) => {
-            return CustomValueStrategy.args(node, customTypes);
+        processArgs: (node, customTypes, boxPrimitive) => {
+            return CustomValueStrategy.args(node, customTypes, boxPrimitive);
         },
         createValue: (args) => {
             return args;
@@ -75,8 +84,8 @@ function createStArrayCustomFunction() {
 
 function createStMapCustomFunction() {
     return createCustomValue<BoxedValueMap, BoxedValueMap>({
-        processArgs: (node, customTypes) => {
-            return CustomValueStrategy.named(node, customTypes);
+        processArgs: (node, customTypes, boxPrimitive) => {
+            return CustomValueStrategy.named(node, customTypes, boxPrimitive);
         },
         createValue: (args) => {
             return args;
@@ -104,7 +113,7 @@ export const deprecatedStFunctions: Record<string, { alternativeName: string }> 
 };
 
 export const CustomValueStrategy = {
-    args: (fnNode: ParsedValue, customTypes: CustomTypes) => {
+    args: (fnNode: ParsedValue, customTypes: CustomTypes, boxPrimitive?: boolean) => {
         const pathArgs = getFormatterArgs(fnNode);
         const outputArray = [];
         for (const arg of pathArgs) {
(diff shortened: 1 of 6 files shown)


packages/core/test/features/st-var.spec.ts
```diff
@@ -1259,162 +1284,569 @@ describe(`features/st-var`, () => {
...
+
+            it('should get deep computed custom value st-var', () => {
+                const { stylable, sheets } = testStylableCore({
+                    '/entry.st.css': `
+                    @st-import [stBorder] from './st-border.js';
+    
+                    :vars {
+                        array: st-array(blue, stBorder(1px, solid, blue));
+                        map: st-map(
+                                border stBorder(value(array, 1, size),
```
idoros, line 1416: like breaks is off here. stBorder first param is inline and the other 2 are on their own lines.

packages/core/test/features/st-var.spec.ts
```diff
@@ -1259,162 +1284,569 @@ describe(`features/st-var`, () => {
...
+                });
+            });
+
+            it('should emit diagnostics only on invalid custom st-vars', () => {
+                const { stylable, sheets } = testStylableCore({
+                    '/entry.st.css': `
+                        @st-import [stBorder] from './st-border.js';
+    
+                        :vars {
+                            border: stBorder(st-array(1px, 2px), solid, red);
```
idoros, line 1586: can't we also add the diagnostic as part of the transformation resolve?

packages/core/src/features/st-var.ts
```diff
@@ -47,8 +54,8 @@ export const diagnostics = {
             .map((s, i) => (i === cyclicChain.length - 1 ? '↻ ' : i === 0 ? '→ ' : '↪ ') + s)
             .join('\n')}"`,
     MISSING_VAR_IN_VALUE: () => `invalid value() with no var identifier`,
-    COULD_NOT_RESOLVE_VALUE: (args: string) =>
-        `cannot resolve value function using the arguments provided: "${args}"`,
+    COULD_NOT_RESOLVE_VALUE: (args?: string) =>
```
idoros, line 57: missing test case that expects any args

packages/core/test/features/st-var.spec.ts
```diff
@@ -1259,162 +1284,569 @@ describe(`features/st-var`, () => {
...
-            const computedVars = stylable.stVar.getComputed(meta);
+                const meta = sheets['/entry.st.css'].meta;
+                const flattenStVars = stylable.stVar.flatten(meta);
 
-            expect(Object.keys(computedVars)).to.eql(['validBefore', 'invalid', 'validAfter']);
-            expect(computedVars.validBefore).to.containSubset({
-                value: 'red',
-                input: undefined,
-                diagnostics: { reports: [] },
+                expect(flattenStVars).to.eql([
```
idoros, line 1753: missing border inner parts paths

packages/core/src/functions.ts
```diff
@@ -30,10 +31,13 @@ export interface EvalValueData {
     tsVarOverride?: Record<string, string> | null;
     cssVarsMapping?: Record<string, string>;
     args?: string[];
+    rootArgument?: string;
+    evaluatorNode?: postcss.Node;
```
idoros, line 35: why is the new `evaluatorNode` used over the already passed down `node` - is `node` overridden at some nested flow?

packages/core/src/features/st-var.ts
```diff
@@ -28,10 +27,18 @@ export interface VarSymbol {
     node: postcss.Node;
 }
 
+export type Input = Box<string, Input | Record<string, Input | string> | Array<Input | string>>;
```
idoros, line 30: maybe `CustomValueInput`

packages/core/src/features/st-var.ts
```diff
@@ -28,10 +27,18 @@ export interface VarSymbol {
...
     value: RuntimeStVar;
     diagnostics: Diagnostics;
-    input?: any;
+    input: Input;
+}
+
+export interface FlatComputedStVar {
+    value: string;
+    symbol: string;
+    path: string[];
```
idoros, line 41: I find `symbol` unclear, why is it needed?

packages/core/src/custom-values.ts
```diff
@@ -3,25 +3,45 @@ import postcssValueParser from 'postcss-value-parser';
 import { getFormatterArgs, getNamedArgs, getStringValue } from './helpers/value';
 import type { ParsedValue } from './types';
 
+export class ValueError extends Error {
```
idoros, line 6: maybe `CustomValueError`?

packages/core/src/custom-values.ts
```diff
@@ -3,25 +3,45 @@ import postcssValueParser from 'postcss-value-parser';
...
+    unboxPrimitives = true,
+    customValues?: CustomTypes,
+    node?: ParsedValue
+): any {
     if (typeof boxed === 'string') {
-        return boxed;
-    } else if (typeof boxed === 'object' && boxed.type && hasOwnProperty.call(boxed, 'value')) {
-        return cloneDeepWith(boxed.value, unbox);
+        return unboxPrimitives ? boxed : box('string', boxed, boxed);
+    } else if (typeof boxed === 'object' && boxed !== null) {
```
idoros, line 38:
```suggestion
    } else if (boxed && typeof boxed === 'object') {
```

packages/core/src/custom-values.ts
```diff
@@ -208,6 +252,8 @@ export function getBoxValue(
         return value;
     } else if (value && customTypes[value.type]) {
         return customTypes[value.type].getValue(path, value, node, customTypes);
+    } else if (value.type === 'string') {
```
idoros, line 255: `string` type value can be overridden by the content of the stylesheet:
```css
@st-import [string] from './custom-vals';
:vars {
  x: string(a, b, c);
}
```

idoros (changes requested):

packages/core/src/features/st-var.ts
tzachbon: We will call `symbol` -> `jsPath`

packages/core/test/features/st-var.spec.ts
```diff
@@ -1259,162 +1284,569 @@ describe(`features/st-var`, () => {
...
+
+                    `
+                );
+
+                const meta = sheets['/entry.st.css'].meta;
+                const flattenStVars = stylable.stVar.flatten(meta);
+
+                expect(flattenStVars).to.eql([
+                    {
+                        symbol: 'nestedArray[0]',
```
tzachbon, line 1780: It's might be problematic to stringify the path in some cases, it should be escaped when needed.

packages/core/src/features/st-var.ts
tzachbon: This isn't a new diagnostics, it is tested there.

packages/core/src/features/st-var.ts
```diff
@@ -286,17 +325,14 @@ function evaluateValueCall(
                     args: restArgs,
                     node: resolvedVarSymbol.node,
                     meta: resolvedVar.meta,
+                    rootArgument: varName,
+                    evaluatorNode: node,
```
tzachbon, line 329: `evaluatorNode` is probably a more fitting name is `initialNode`.
In future refactor it should on the context (currently add a "TODO" comment)

packages/core/src/functions.ts
tzachbon: `evaluatorNode` - probably a more fitting name is `initialNode` and it's used to give initial node as context.

packages/core/test/features/st-var.spec.ts
tzachbon: Add a skipped test that will check this problem, it will be implemented in a new PR.

packages/core/test/features/st-var.spec.ts
tzachbon, line 1780: We will remove the `symbol`

packages/core/src/custom-values.ts
tzachbon: https://github.com/wix/stylable/issues/2424

packages/core/src/custom-values.ts
```diff
@@ -252,7 +257,7 @@ export function getBoxValue(
         return value;
     } else if (value && customTypes[value.type]) {
         return customTypes[value.type].getValue(path, value, node, customTypes);
-    } else if (value.type === 'string') {
+    } else if (value.type === 'st-string') {
         return (value as Box<'string', string>).value;
```
idoros, line 261: shouldn't it create a `st-box` type or use the `boxString` function?

tzachbon: It should unbox it, I changed it to use `unbox`

idoros (approved):

PR #2426, opened 2022-04-04
added ability to dedupe duplicated modules

barak007:


packages/webpack-plugin/src/plugin.ts
```diff
@@ -51,7 +51,16 @@ import { parse } from 'postcss';
...
+    /**
+     * Enable css minification
+     */
     minify?: boolean;
+    /**
+     * @experimental
+     * Remove modules with the same target css and same depth from output
+     * This can be beneficial when you have multiple versions of the same package in your project 
+     */
+    dedupeSimilarStylesheets?: boolean;
```
barak007, line 63: rename `dedupeSimilarStylesheets`

PR #2430, opened 2022-04-07
feat(create-stylable-app): add `stcConfig` to the integrations templates

tzachbon:
After we introduced the option you use `stc` in our integration https://github.com/wix/stylable/pull/2342, this PR adds this ability to our templates.
We can now generate Stylable stylesheets declaration files to give the users a strong typed experience when using Stylable and Typescript together seamlessly.

> This process requires the users to "run" their integration while developing.

- [x] Add documentation to `stylable.io` https://github.com/wixplosives/stylable.io/pull/35

### DIFF
--- a/packages/create-stylable-app/template/ts-react-webpack/template.js
+++ b/packages/create-stylable-app/template/ts-react-webpack/template.js
@@ -1,8 +1,10 @@
+//@ts-check
 /** @type {import('create-stylable-app').TemplateDefinition} */
 module.exports = {
     dependencies: ['react', 'react-dom'],
     devDependencies: [
         '@stylable/core',
+        '@stylable/cli',
         '@stylable/runtime',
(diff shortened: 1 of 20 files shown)


packages/create-stylable-app/template/ts-react-webpack/template.js
```diff
@@ -38,4 +40,5 @@ module.exports = {
             test: 'npm run typecheck && npm run lint',
         },
     },
+    postinstall: [['npm', 'run', 'build']],
```
AviVahl, line 43: `postinstall` runs when installing an npm package as a dep. if anything, use `prepare`, but I'm not sure this is even a wanted behavior...

tzachbon: This is our "postinstall", it's not added as a script to the user's package json

PR #2440, opened 2022-04-12
feat(core): expose meta dependencies API

tzachbon:
## Expose meta dependencies API (BREAKING)

> Should be merged to `ido/stylable-v5`


barak007:
I like it....where is the breaking change? the visit.....?

tzachbon:
> I like it....where is the breaking change? the visit.....?

removing the visit from the index, and its arguments.

packages/core/src/visit-meta-css-dependencies.ts
```diff
@@ -2,12 +2,29 @@ import type { StylableMeta } from './stylable-meta';
...
+    meta: StylableMeta;
+}
+
+export interface JSDependency {
+    kind: 'js';
+    resolvedPath: string;
+    imported: Imported;
+}
+
+export type MetaDependency = CSSDependency | JSDependency;
```
idoros, line 19:
```suggestion
export type Dependency = CSSDependency | JSDependency;
```

PR #2441, opened 2022-04-12
feat(cli): remove default `cjs` format

### LINKED ISSUE #1854: CLI should have no output module type defined by default
Currently when running `stc` the only module type that defaults to true is `cjs`. While this can fit some use-cases, it definitely doesn't fit all of them.

In the next major version, we can break this behavior and force the user to specify all module types they require.

Something like:
```sh
stc --modules="cjs, stcss, dts, dtsSourceMap"
```

tzachbon:
## BREAKING

closes #1854

### DIFF
--- a/packages/cli/src/build-stylable.ts
+++ b/packages/cli/src/build-stylable.ts
@@ -83,7 +83,10 @@ export async function buildStylable(
             log('[Project]', projectRoot, buildOptions);
 
             if (!hasStylableCSSOutput(buildOptions)) {
-                log(`No CSS output found for "${identifier}"`, levels.info);
+                log(
+                    `No target output declared "${identifier}", please provide one or more of the following target options: "cjs", "esm", "css", "stcss" or "indexFile"`,
+                    levels.info
+                );
             }
 
             const stylable = new Stylable({
--- a/packages/cli/src/build-stylable.ts
+++ b/packages/cli/src/build-stylable.ts
@@ -84,7 +84,7 @@ export async function buildStylable(
 
             if (!hasStylableCSSOutput(buildOptions)) {
                 log(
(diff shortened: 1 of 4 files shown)


packages/cli/src/build-stylable.ts
```diff
@@ -81,6 +82,10 @@ export async function buildStylable(
 
             log('[Project]', projectRoot, buildOptions);
 
+            if (!hasStylableCSSOutput(buildOptions)) {
+                log(`No CSS output found for "${identifier}"`, levels.info);
```
tomrav, line 86: Perhaps consider changing to something like:
`No target output declared, please provide one or more of the following target options: "cjs", "esm", "css", "stcss" or "indexFile"`

tomrav: and possibly a higher severity?

barak007:
There is another change we can make for the cli interface and that's to be able to emit output and also build index I the same command. Right now if you add indexFile it will not emit assets.

tzachbon:
> There is another change we can make for the cli interface and that's to be able to emit output and also build index I the same command. Right now if you add indexFile it will not emit assets. Maybe should be added to the v5 issue

Agreed, I'll add it to the issue and we will change it in a different PR.

packages/cli/src/build-stylable.ts
```diff
@@ -81,6 +82,13 @@ export async function buildStylable(
 
             log('[Project]', projectRoot, buildOptions);
 
+            if (!hasStylableCSSOutput(buildOptions)) {
+                log(
+                    `No target output declared "${identifier}", please provide one or more of the following target options: "cjs", "esm", "css", "stcss" or "indexFile"`,
```
tomrav, line 87:
```suggestion
                    `No target output declared for "${identifier}", please provide one or more of the following target options: "cjs", "esm", "css", "stcss" or "indexFile"`,
```

tomrav (approved):

PR #2448, opened 2022-04-13
bug: fix issue when bundling with target "node" 

barak007:


packages/core/src/stylable-resolver.ts
```diff
@@ -108,19 +108,18 @@ export class StylableResolver {
         }
 
         let entity: CachedModuleEntity;
+        const resolvedPath = this.resolvePath(context, request);
 
-        if (request.endsWith('.css')) {
+        if (resolvedPath.endsWith('.css')) {
```
tzachbon, line 113: If the resolution fails it is not included in the "catch".

barak007: should be not finished yet and I'm thinking what actually do here. because we don't know the kind

PR #2469, opened 2022-04-25
feat(cli): generate index file with other outputs; `IndexGenerator` refactor

### LINKED ISSUE #2470: Stylable CLI: be able to create index file and other output formats in the same process.
Currently, this feature is unavailable. You either create an index file or other file formats.

tzachbon:
# Changes

* It's possible to create an index file with additional output file formats.
* The index file will target the output st.css if it exists.
* `Base` symbol removed from the CLI index file.
* `IndexGenerator` API parameters changes.

closes #2470

### DIFF
--- a/packages/cli/test/generate-index.spec.ts
+++ b/packages/cli/test/generate-index.spec.ts
@@ -96,6 +96,52 @@ describe('build index', () => {
             ].join('\n')
         );
     });
+
+    it('should create index file importing all matched stylesheets in srcDir when has output cjs files', async () => {
+        const fs = createMemoryFs({
+            src: {
+                '/compA.st.css': `
+                .a{}
+                `,
+                '/a/b/comp-B.st.css': `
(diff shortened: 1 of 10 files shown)


packages/cli/test/generate-index.spec.ts
```diff
@@ -51,6 +51,51 @@ describe('build index', () => {
...
+            fileSystem: fs,
+            requireModule: () => ({}),
+        });
+
+        await build(
+            {
+                outDir: './dist',
+                srcDir: './src',
+                indexFile: 'index.st.css',
+                outputSources: true,
```
barak007, line 77: we need test when `outputSources` is false and we emit only js files

tzachbon: we do not emit `js` files by default (in this branch)

barak007 (approved):

PR #2484, opened 2022-04-29
fix: collect `meta.mixins` only during analyze

idoros:
This PR fixes a regression that was introduced in the mixins analyze data. The `meta.mixins`, originally holds the list of all mixins in a stylesheet, was wrongfully pushed more definitions during the transform phase.

Notice that `meta.mixins` is deprecated and will be removed soon in `v5`

### DIFF
--- a/packages/core/src/features/st-mixin.ts
+++ b/packages/core/src/features/st-mixin.ts
@@ -146,7 +146,7 @@ function collectDeclMixins(
     context: FeatureContext,
     decl: postcss.Declaration,
     paramSignature: (mixinSymbolName: string) => 'named' | 'args',
-    emitStrategyDiagnostics: boolean,
+    isTransformPhase: boolean,
     previousMixins?: RefedMixin[]
 ): RefedMixin[] {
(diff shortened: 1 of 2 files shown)


tomrav (approved):

packages/core/src/features/st-mixin.ts
```diff
@@ -161,7 +161,7 @@ function collectDeclMixins(
         return previousMixins || mixins;
     }
 
-    parser(decl, paramSignature, context.diagnostics, emitStrategyDiagnostics).forEach((mixin) => {
+    parser(decl, paramSignature, context.diagnostics, isTransformPhase).forEach((mixin) => {
```
barak007, line 164: was `emitStrategyDiagnostics` unused?

idoros: not unused. it causes the parser to output diagnostics related to params parsing. I connected both the emit and the push of mixins to meta into one as this is only relevant to transformation phase and only a short solution for `v4`. in our next major this code is all gone.

PR #2489, opened 2022-05-03
refactor(core): prepare process AST mods for move to transform

idoros:
This PR refactor the AST modifications that occur in the processor to prepare them to be moved to the transformer

- [x] `prepareAST` - new feature hook called on every node at the end of the analyze process
- modifiers
  - [x] `@namespace`
  - [x] `:vars`
  - [x] `@st-import` / `:import`
  - [x] `@st-global-custom-property`
  - [x] `@custom-selector`
  - [x] `@st-scope`

### DIFF
--- a/packages/core/src/features/st-scope.ts
+++ b/packages/core/src/features/st-scope.ts
@@ -1,0 +1,59 @@
+import { createFeature } from './feature';
+import { parseSelectorWithCache, scopeNestedSelector } from '../helpers/selector';
+import type { ImmutablePseudoClass } from '@tokey/css-selector-parser';
+import * as postcss from 'postcss';
+import type { SRule } from '../deprecated/postcss-ast-extension';
+
+export const diagnostics = {
+    MISSING_SCOPING_PARAM() {
+        return '"@st-scope" missing scoping selector parameter';
+    },
+};
+
+// HOOKS
+
+export const hooks = createFeature<{ IMMUTABLE_SELECTOR: ImmutablePseudoClass }>({
+    analyzeAtRule({ context, atRule, analyzeRule }) {
+        if (!isStScopeStatement(atRule)) {
+            return;
+        }
+        if (!atRule.params) {
+            context.diagnostics.warn(atRule, diagnostics.MISSING_SCOPING_PARAM());
+        }
+        analyzeRule(
+            postcss.rule({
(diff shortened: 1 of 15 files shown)


packages/core/src/features/st-scope.ts
```diff
@@ -0,0 +1,59 @@
...
+            {
+                isScoped: true,
+            }
+        );
+        context.meta.scopes.push(atRule);
+    },
+    prepareAST({ node, toRemove }) {
+        if (isStScopeStatement(node)) {
+            flattenScope(node);
+            toRemove.push(() => node.replaceWith(node.nodes || []));
```
barak007, line 37: maybe we want to change the name toRemove this is doing the replacement

packages/core/src/stylable-processor.ts
```diff
@@ -181,33 +199,37 @@ export class StylableProcessor implements FeatureContext {
                     CSSKeyframes.hooks.analyzeAtRule({
                         context: this,
                         atRule,
-                        toRemove,
+                        analyzeRule,
                     });
                     break;
                 case 'custom-selector': {
                     const params = atRule.params.split(/\s/);
```
barak007, line 206: do we want to extract custom-selector to `analyzeAtRule`

packages/core/src/stylable-processor.ts
```diff
@@ -287,6 +313,11 @@ export class StylableProcessor implements FeatureContext {
                         rule,
                         walkContext: nodeContext,
                     });
+                } else if (node.value.startsWith('--')) {
```
barak007, line 316: we have a function for this check

packages/core/src/features/st-scope.ts
idoros: not sure of what's a better name for this, anyhow it should be dismantled into the transformer flow in the future, so the API should be removed eventually

packages/core/src/stylable-processor.ts
idoros: not in this PR, feature extraction should done in a separate PR

packages/core/src/stylable-processor.ts
idoros: we don't. there is a similar function that validates the name of a custom property. nothing like that for custom selector

PR #2515, opened 2022-05-17
fix(cli): add undeclared dependency "@stylable/module-utils"

jakutis:


tomrav (changes requested):
Thanks for opening the PR.
Please run `npm i` in the root to update the lock file as well.

barak007:
Nice catch.
@tomrav we probably want the monorepo validation tools running on all Stylable packages.

tomrav:
I added this dependency while I was doing upgrades. Thanks for opening the PR.

PR #2516, opened 2022-05-17
feat: add ability to pass buildNamespace function to default namespace strategy

barak007:
This PR will allow others to use our default namespace strategy with more complex namespace logic for each package
This will open the possibility to enable `module federation` and allow the federated packages to get different consistent namespace then the rest of the build

### DIFF
--- a/packages/core/src/index-deprecated.ts
+++ b/packages/core/src/index-deprecated.ts
@@ -159,7 +159,11 @@ export {
     nativePseudoClasses,
     nativePseudoElements,
 } from './native-reserved-lists';
-export { noCollisionNamespace, packageNamespaceFactory } from './resolve-namespace-factories';
+export {
+    noCollisionNamespace,
+    packageNamespaceFactory,
+    defaultBuildNamespace,
+} from './resolve-namespace-factories';
 export { DiagnosticsMode, EmitDiagnosticsContext, emitDiagnostics } from './report-diagnostic';
 import { visitMetaCSSDependenciesBFS as deprecatedVisitMetaCSSDependenciesBFS } from './visit-meta-css-dependencies';
 /**@deprecated use Stylable.getDependencies in v5*/
--- a/packages/core/src/resolve-namespace-factories.ts
+++ b/packages/core/src/resolve-namespace-factories.ts
@@ -10,7 +10,8 @@ export function packageNamespaceFactory(
     }: { dirname(path: string): string; relative(from: string, to: string): string },
     hashSalt = '',
     prefix = '',
-    normalizeVersion = (semver: string) => semver
+    normalizeVersion = (semver: string, _packageName: string) => semver,
+    buildNamespace = defaultBuildNamespace
 ): typeof processNamespace {
     return (namespace: string, stylesheetPath: string) => {
         const configPath = findConfig('package.json', { cwd: dirname(stylesheetPath) });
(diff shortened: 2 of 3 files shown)


packages/core/src/resolve-namespace-factories.ts
```diff
@@ -10,25 +10,43 @@ export function packageNamespaceFactory(
     }: { dirname(path: string): string; relative(from: string, to: string): string },
     hashSalt = '',
     prefix = '',
-    normalizeVersion = (semver: string) => semver
+    normalizeVersion = (semver: string, _packageName: string) => semver,
```
barak007, line 13: this was a mistake in the API it was not flexible enough.

barak007:
One more thing to consider is to pass the hash function to buildNamespace as parameter or export it properly from stylable

packages/core/src/resolve-namespace-factories.ts
```diff
@@ -10,25 +10,43 @@ export function packageNamespaceFactory(
...
         const config = loadConfig(configPath) as { name: string; version: string };
-        const fromRoot = relative(dirname(configPath), stylesheetPath).replace(/\\/g, '/');
-        return (
-            prefix +
-            namespace +
-            murmurhash3_32_gc(
-                hashSalt + config.name + '@' + normalizeVersion(config.version) + '/' + fromRoot
-            )
+        return buildNamespace(
+            hashSalt,
```
johnbenz13, line 23: `hashSalt` value should be dependent on `config` in line `21` otherwise the `namespace` will be still linked to the artifactId (see the change I had to do [here](https://github.com/johnbenz13/mf-with-css-encapsulation/pull/3/files#diff-d1ccad9baa23796c3419806748865da9241ed48b4c3983945c1a1c6eb294ac9cR53))
so we could make it a function as well, something like:

```
hashSalt: (config): (config: {name: string; version: string}) => string
```

barak007: The buildNamespace function can ignore this salt same way you did base on the package name so this still enable the same transformation. Also you will not use the normalizeVersion function and the entire logic can sit in the same place

packages/core/src/resolve-namespace-factories.ts
```diff
@@ -10,25 +10,43 @@ export function packageNamespaceFactory(
...
+    prefix: string,
+    namespace: string,
+    packageName: string,
+    packageVersion: string,
+    relativeFilePathFromPackageRoot: string
+) {
+    return (
+        prefix +
+        namespace +
+        murmurhash3_32_gc(
```
johnbenz13, line 44: what about `murmurhash3_32_gc`? I'm not sure why I should use it but if it is important it would be good to have a way to reuse the same function you're using

barak007: Wrote about exporting it or pass it as an argument. You will have access to it in one way or another

packages/core/src/index-deprecated.ts
```diff
@@ -159,7 +159,11 @@ export {
     nativePseudoClasses,
     nativePseudoElements,
 } from './native-reserved-lists';
-export { noCollisionNamespace, packageNamespaceFactory } from './resolve-namespace-factories';
+export {
+    noCollisionNamespace,
+    packageNamespaceFactory,
+    defaultBuildNamespace,
+} from './resolve-namespace-factories';
```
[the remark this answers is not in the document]
idoros: `noCollisionNamespace` is exported, the fact that it appears here is a mistake, and `packageNamespaceFactory` wasn't in the known used list.

We can export them in v5 (`noCollisionNamespace` is already there). It would be nice to have them under identical prefix, "namespace", maybe?

packages/core/src/resolve-namespace-factories.ts
```diff
@@ -10,25 +10,43 @@ export function packageNamespaceFactory(
...
 }
 
+export function defaultBuildNamespace(
+    hashSalt: string,
+    prefix: string,
+    namespace: string,
+    packageName: string,
+    packageVersion: string,
+    relativeFilePathFromPackageRoot: string
+) {
```
idoros, line 40: maybe change to named options?

tomrav:
Closing this in favor of #2523

PR #2526, opened 2022-05-19
fix: cachedProcessFile should not based on mtime

barak007:
After noticing mtime based cache can lead to unexpected false negative.
This PR removes the mtime based caching in the cachedProcessFile


idoros (approved):
Looks good! Now we how many times do you want to run it?

PR #2531, opened 2022-05-23
fix(language-service): filter out false diagnostics for st-scope selectors with special characters

### LINKED ISSUE #2200: LSP issues false diagnostic about `@stScope` params containing special charactars (:, [], *...)
### Actual Behavior

The example below works in practice, but the LSP wrongly issues a warning `media query expected `.
This error originates from the `css-language-service` we use and should be filtered out like we do in other cases.

```css
@st-scope [dir=rtl] {
    .root {}
}
```

Update: a similar sort of error happens with scoping usage of `:` and `*`. Looks like we need to sanitize the entirety of `@st-scope` parameters, and not just the partial class handling present there now.

### Expected Behavior

Should not issue a diagnostic for this case.

### Version

`@stylable/language-service@4.9.1` (latest)

tomrav:
Resolves #2200

### DIFF
--- a/packages/core/src/helpers/ast.ts
+++ b/packages/core/src/helpers/ast.ts
@@ -1,0 +1,26 @@
+import type * as postcss from 'postcss';
+
+// ToDo check empty file (position 0)
+export function getAstNodeAt(
+    root: postcss.Root,
+    offset: number
+): [node: postcss.AnyNode, offsetInNode: number] {
+    let closestNode: postcss.AnyNode = root;
+    let offsetInNode = 0;
+
+    root.walk((node) => {
+        if (node.source?.start && node.source?.end) {
+            const inNode =
+                node.source.start.offset <= offset && node.source.end.offset + 1 >= offset;
+            if (!inNode) {
+                // out of node: bailout
+                return false;
+            }
+            closestNode = node;
+            offsetInNode = offset - node.source.start.offset;
+        }
+        return;
+    });
+
+    return [closestNode, offsetInNode];
+}
--- a/packages/core/test/helpers/ast.spec.ts
+++ b/packages/core/test/helpers/ast.spec.ts
@@ -1,0 +1,168 @@
+import { expect } from 'chai';
+import { parse } from 'postcss';
+import { getAstNodeAt } from '@stylable/core/dist/helpers/ast';
+import deindent from 'deindent';
+
+function setupWithCursor(source: string) {
+    const deindented = deindent(source);
+
+    const position = deindented.indexOf(`|`);
+    return {
+        position,
+        ast: parse(deindented.split(`|`).join(``)),
+    };
+}
+
+// ToDo: make added syntax zero space: semicolon at declaration end
+describe('helpers/ast', () => {
+    describe(`getAstNodeAt`, () => {
+        it(`should find position in declaration property (start)`, () => {
+            const { position, ast } = setupWithCursor(`
+                .selector {
+                    |decl: declValue;
+                }
+            `);
+
+            const [node, offsetInNode] = getAstNodeAt(ast, position);
+
+            expect(node, 'node').to.equal((ast as any).nodes[0].nodes[0]);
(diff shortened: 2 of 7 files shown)


packages/language-service/test/lib/diagnostics.spec.ts
```diff
@@ -169,5 +170,39 @@ describe('diagnostics', () => {
 
             expect(diagnostics).to.eql([]);
         });
+
+        it('should ignore errors for attribute selectors inside @st-scope', () => {
+            const filePath = '/style.st.css';
```
idoros, line 175: Unify the two tests, you can test both cases at once

packages/language-service/test/lib/diagnostics.spec.ts
```diff
@@ -169,5 +170,39 @@ describe('diagnostics', () => {
 
             expect(diagnostics).to.eql([]);
         });
+
+        it('should ignore errors for attribute selectors inside @st-scope', () => {
```
idoros, line 174:
```suggestion
        it('should ignore native css lsp errors for attribute selectors inside @st-scope', () => {
```

packages/language-service/src/lib/css-service.ts
```diff
@@ -128,59 +129,85 @@ export class CssService {
...
                         return false;
                     }
+                } else if (
+                    diag.code === 'css-mediaqueryexpected' ||
+                    diag.code === 'css-ruleorselectorexpected'
+                ) {
+                    // caused by @st-scope with special characters (e.g. @st-scope [div=rtl])
+                    const offset = getOffsetFromPosition(meta.rawAst.toString(), diag.range.start);
+                    const [astNode, _astNodeOffset] = getAstNodeAt(meta.rawAst, offset + 1);
+                    const stScopeSanitized = 'media   ';
```
idoros, line 180: We have to get rid of the native css lsp

idoros (approved):

packages/core/src/helpers/ast.ts
```diff
@@ -0,0 +1,26 @@
...
+// ToDo check empty file (position 0)
+export function getAstNodeAt(
+    root: postcss.Root,
+    offset: number
+): [node: postcss.AnyNode, offsetInNode: number] {
+    let closestNode: postcss.AnyNode = root;
+    let offsetInNode = 0;
+
+    root.walk((node) => {
+        if (node.source?.start && node.source?.end) {
```
barak007, line 12: Do we care about zeros? What is this check for see that you have source?

packages/core/test/helpers/ast.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+            `);
+
+            const [node, offsetInNode] = getAstNodeAt(ast, position);
+
+            expect(node, 'node').to.equal((ast as any).nodes[0]);
+            expect(offsetInNode, 'offset').to.equal(15);
+        });
+    });
+
+    xdescribe(`parseForInspection (safe parse)`, () => {
```
barak007, line 81: Remove?

packages/language-service/test/lib/diagnostics.spec.ts
tomrav: done

packages/language-service/test/lib/diagnostics.spec.ts
tomrav: done

packages/core/test/helpers/ast.spec.ts
tomrav: we left these here intentionally for future work.

packages/language-service/src/lib/css-service.ts
tomrav: I knoooow.

packages/core/src/helpers/ast.ts
tomrav: `start` and `end` are objects and cannot be 0 (they contain, line, column and offset).

PR #2568, opened 2022-06-19
refactor: avoid calling stat manually on each entry

barak007:
Big speedup on large file sets.


packages/node/src/find-files.ts
```diff
@@ -17,16 +17,15 @@ export function findFiles(
     while (folders.length) {
         const current = folders.pop()!;
         try {
-            fs.readdirSync(current).forEach((item: string) => {
-                if (blacklist.has(item)) {
+            fs.readdirSync(current, { withFileTypes: true }).forEach((item) => {
```
AviVahl, line 20: should be noted that `withFileTypes` behaves like `lstat`, rather than `stat`.
meaning any symlink will have `isSymbolicLink() === true` and `isFile/Directory() === false`
not sure whether stylable cares about it

barak007: So this is breaks behavior whenever there is a symbolic link.
For my case of code mods it was better not to handle links. 
After looking at usages in Stylable I think it's ok to change this behavior. It's only used twice once in the context of code-mods and once in the metadata-plugin both cases are looking for "real files"

WDYT?

barak007: @AviVahl

AviVahl: usually the safety around links (being explicitly marked) is good for code. allows not to get into recursive loops if u don't protect visited paths.

also make sure to remove the try-catch where the statSync call was. unneeded now

idoros (approved):

PR #2577, opened 2022-06-27
fix: report only top unknown argument for top mixin

idoros:
This PR fixes a diagnostic report that leaks out of transformation of internal mixins. Transformation context now have a `topTransform` field that contains the initial meta that started the transformation so diagnostics can be reported accordingly.

## **update:**
After discussing this in person some more, it might be a mistake to prevent "inlined" errors, as these errors show that the code inlined from another place has issues that might cause unintended behavior. That being said, looking at other diagnostics like this, some are exposed, others aren't, and some are not reported correctly and should be fixed: [playground](https://wixplosives.github.io/stylable-playground2/?filePath=%2Fsrc%2Fapp.st.css&projectId=Om1b0SJC83R5sbrUH1cl&type=code) to demonstrate some cases.


idoros:
closing for now

PR #2579, opened 2022-06-27
refactor!: simplify `stylable.transform` API (v5)

idoros:
This PR simplify the `stylable.transform`:

## simplify input
Change signature from accepting either `meta or source+path` to accepting `meta or path`. The new signature is inline with other top level transform functions like `transformSelector` or `transformDecl`. 

To transform custom source on specific context path now you first need to create a `meta` and pass it to the transform method: `stylable.transform(stylable.analyze(path, source), {/*optional transform options*/})`

This also fix a Typescript issue that the old signature could theoretically accept transform options, but couldn't in reality since they weren't in the typescript signature.

## remove cache on transform

Calling transform used to replace the file processor cache with whatever was passed in. That means that potential override content that was not synced with filesystem would be returned for future calls. This PR changes transform so it doesn't directly change the cache. Only calls to `stylable.analyze(path)` (without override content) caches, and calling `stylable.transform(path)` calls `analyze`.

### DIFF
--- a/packages/core/src/stylable.ts
+++ b/packages/core/src/stylable.ts
@@ -146,21 +146,11 @@ export class Stylable {
             ...options,
         });
     }
-    // ToDo: unify signature - accept only meta + optional transformer options
-    public transform(meta: StylableMeta): StylableResults;
-    public transform(source: string, resourcePath: string): StylableResults;
     public transform(
-        metaOrSource: string | StylableMeta,
-        resourcePath?: string,
-        options: Partial<TransformerOptions> = {},
-        processorOptions: CreateProcessorOptions = {}
+        pathOrMeta: string | StylableMeta,
+        options: Partial<TransformerOptions> = {}
     ): StylableResults {
-        const meta =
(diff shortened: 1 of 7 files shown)


packages/core/src/stylable.ts
```diff
@@ -146,21 +146,11 @@ export class Stylable {
...
     ): StylableResults {
-        const meta =
-            typeof metaOrSource === 'string'
-                ? this.createProcessor(processorOptions).process(
-                      this.cssParser(metaOrSource, { from: resourcePath })
-                  )
-                : metaOrSource;
+        const meta = typeof pathOrMeta === `string` ? this.analyze(pathOrMeta) : pathOrMeta;
         const transformer = this.createTransformer(options);
         this.fileProcessor.add(meta.source, meta);
```
barak007, line 155: can be removed since we use analyze and it does cache

idoros: removed the cache in transform. had to fix CLI code that "tried" to optimize by overriding content with identical content of the filesystem.

PR #2580, opened 2022-06-28
feat(core): public API to get st-scope atrule

idoros:
This PR adds an API to find the `@st-scope` atrule of a nested rule from source AST. The API is available on stylable instance:

```js
const stScope = stylable.stScope.getStScope(rule);
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -37,3 +37,4 @@ export { createDefaultResolver } from './module-resolver';
 // low-level api
 export { parseModuleImportStatement, ensureModuleImport } from './helpers/import';
 export { validateCustomPropertyName } from './helpers/css-custom-property';
+export { getStScope } from './features/st-scope';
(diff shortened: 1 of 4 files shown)


packages/core/src/index.ts
```diff
@@ -37,3 +37,4 @@ export { createDefaultResolver } from './module-resolver';
 // low-level api
 export { parseModuleImportStatement, ensureModuleImport } from './helpers/import';
 export { validateCustomPropertyName } from './helpers/css-custom-property';
+export { getStScope } from './features/st-scope';
```
barak007, line 40: why export if we have it on stylable?

idoros: I prefer not to expose low level APIs at this point, I thought you wanted it as an alternative

barak007: from what I saw in the code we have stylable instance very close to the usage. we can probably only expose the API

idoros: removed

PR #2597, opened 2022-07-07
fix(integrations): collect dependencies resiliently

barak007:
This PR fixes the collection of dependencies to not rely on parsing the dependent file.
Along the way, we are now using the `resolverCache` for the public `resolvePath` of `Stylable` and `StylableResover`.
This might be risky and we need to think if it's not.


packages/build-tools/src/try-collect-imports-deep.ts
```diff
@@ -0,0 +1,18 @@
...
+    meta: StylableMeta,
+    imports = new Set<string>()
+) {
+    for (const { context, request } of meta.getImportStatements()) {
+        try {
+            const resolved = stylable.resolver.resolvePath(context, request);
+            imports.add(resolved);
+            tryCollectImportsDeep(stylable, stylable.analyze(resolved), imports);
+        } catch (e) {
+            /** */
```
tomrav, line 14: need to report failures back out

packages/build-tools/src/try-collect-imports-deep.ts
```diff
@@ -0,0 +1,18 @@
+import type { Stylable, StylableMeta } from '@stylable/core';
+
+export function tryCollectImportsDeep(
```
tomrav, line 3: seems like this could be a part of `core`, probably in a helper/import with `visitMetaCSSDependencies`

packages/core/src/stylable-resolver.ts
```diff
@@ -32,6 +32,7 @@ export interface InvalidCachedModule {
     error: unknown;
     request: string;
     context: string;
+    resolvedPath: string | undefined;
 }
 
 export interface CachedStylableMeta {
```
tomrav, line 38: is this exposed publicly? wouldn't changes here be considered breaking?

packages/core/src/stylable.ts
```diff
@@ -224,4 +224,7 @@ export class Stylable {
             ? this.fileProcessor.processContent(overrideSrc, fullPath)
             : this.fileProcessor.process(fullPath);
     }
+    public resolvePath(directoryPath: string, request: string) {
```
tomrav, line 227: why is this a top level api? would `stylable.resolver.resolvePath` be better?

packages/experimental-loader/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "@stylable/core": "^4.13.3",
+    "@stylable/build-tools": "^4.13.3",
```
tomrav, line 10: if `tryCollectImportsDeep` moves to `core`, this becomes redundant

packages/build-tools/src/try-collect-imports-deep.ts
tomrav: maybe in `v5` this should be handled via `stylable.getDependencies`, adding resiliency and error reporting

tomrav (changes requested):
left some thoughts.

tomrav (approved):

PR #2609, opened 2022-07-13
feat(create-app): native esm templates

AviVahl:
- use `"type": "module"` in all templates
- prettier now formats cjs/mjs/cts/mts as well
- rename `stylable.config.js` to `stylable.config.cjs`
- remove `"cjs": false` from stylable config, as it's `false` by default
- rename `rollup.config.cjs` to `rollup.config.js`
- fix rollup `sourcemap` warning
- convert webpack and rollup configs to esm


tomrav (approved):

idoros:
Doesn't type module has issues with our generated d.ts files? Maybe we should wait until we research the issue some more.

PR #2624, opened 2022-07-20
feat(core): transition from native `@namespace` (conflict) 

idoros:
This PR adds an alternative to `@namespace` in order to remove conflict with native CSS. 

## Tasks
1. [x] move code to `st-namespace` feature
2. [x] add `@st-namespace` atrule
3. [x] namespace validation:
     - valid (ident-like):
         `@st-namespace "dropdown";`
         `@st-namespace 'dropdown';`
         `@st-namespace "group-card";`
         `@st-namespace "MixCapitalization";`
     - invalid:
         `@st-namespace "http://www.w3.org/1999/xhtml";` <- **collect just ident-like**
         `@st-namespace prefix "http://www.site.com/foo";` <- **no prefix**
         `@st-namespace url(https://www.site.com);`             <- **no url**
4. [x] collect from `@namespace`  for backward compatibility (just when matches validation)
5. [x] remove native `@namespace` **only** if a namespace was generated from it
6. [x] prefer `@st-namespace`
7. [x] change language server to complete `@st-namespace` instead and add it as a known at-rule
8. [x] `@namespace` to `@st-namespace` code mod
- [x] add to v4.x - #2641 
- [ ] update documentation to refer to `@st-namespace` only (maybe a small section about previous behavior)
- [ ] add issue to stop collecting stylesheet namespace from `@namespace` in the next major


packages/core/src/features/st-namespace.ts
```diff
@@ -0,0 +1,199 @@
...
+        return;
+    }
+    if (namespace !== undefined && namespace.trim() === '') {
+        // empty namespace found
+        diag?.report(diagnostics.EMPTY_NAMESPACE_DEF(), {
+            node,
+        });
+        return;
+    }
+    if (!namespace.match(/^[a-zA-Z-]+$/)) {
```
barak007, line 152: add unicode range for css

barak007: add number rage also

barak007: after first char

packages/language-service/src/lib/css-service.ts
```diff
@@ -136,6 +136,7 @@ export class CssService {
                     diag.code === 'unknownAtRules' &&
                     (atRuleName === '@custom-selector' ||
                         atRuleName === '@st-scope' ||
+                        atRuleName === '@st-namespace' ||
```
barak007, line 139: might need a change in `stylable.tmLanguage.json` stylable-intelligence

idoros: I think `stylable.tmLanguage.json` is just for coloring and should fallback to regular at-rule, anyhow not part of this PR

barak007 (approved):

PR #2628, opened 2022-07-24
fix(cli): resolve namespaceResolver path relative to rootDir

barak007:
support relative paths for `namespaceResolver` option for when using cli interface.


idoros (changes requested):
add test to demonstrate the resolved issue

idoros (approved):

PR #2641, opened 2022-08-02
feat(core) transition from native `@namespace` - v4

idoros:
This PR adds support for `@st-namespace` that was added in v5 to v4

- added to v5 in #2624

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -338,6 +338,8 @@ npx -p @stylable/cli stc-codemod --help
 
 - `st-global-custom-property-to-at-property` - Convert deprecated `@st-global-custom-property *;` to `@property st-global(*);` syntax.
 
+- `namespace-to-st-namespace` - Converts `@namespace` that would have been used as Stylable namespace configuration to `@st-namespace`.
+
 ### Provide an external codemod
 
 Codemods are transformation operations for code based on AST.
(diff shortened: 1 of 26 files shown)


packages/core/src/stylable-processor.ts
```diff
@@ -499,10 +458,6 @@ export function createEmptyMeta(root: postcss.Root, diagnostics: Diagnostics): S
     return new StylableMeta(root, diagnostics);
 }
 
-export function processNamespace(namespace: string, origin: string, _source?: string) {
```
barak007, line 502: this used to be exported...after the move is it?

packages/cli/README.md
```diff
@@ -154,7 +154,7 @@ export class Generator extends Base {
     }    
     protected generateIndexSource(indexFileTargetPath: string) {
         const source = super.generateIndexSource(indexFileTargetPath);
-        return '@namespace "INDEX";\n' + source;
+        return '@st-namespace "INDEX";\n' + source;
```
barak007, line 157: I don't think we want this change in v4. it means that any one would must have the latest version...this is kind of breaking...

barak007: it's only a docs example...so I'm not sure it matters

idoros: that's what I thought, it doesn't have any affect on lib usage

PR #2653, opened 2022-08-08
fix(core): fix resolver performance regression (v4)

barak007:
Fix regression that caused Stylable to hit the underlying resolver more then once.


idoros (changes requested):
looks good, lets just add a test to validate caching won't have another regression

idoros (approved):

PR #2668, opened 2022-08-14
fix(core): allow value reference in st-map

idoros:
This PR fixes a bug that prevented nesting referenced `value()` within a `st-map` var type:

```css
:vars {
  list: st-array(red, green);
  map: st-map(key value(list))
}
.a {
  prop: value(map, key, 1) /* should return "green" */

  /* before fix it evaluated to */
  prop: st-array(red, green);
}
``` 

-----

In addition this also fixes the output from `stylable.stVar.getComputed` to return the unboxed value on the input field instead of the runtime value:

```css
:vars {
  green: green;
  inline: st-map(a red);
  reference: st-map(a value(red));
}
```
So that `stylable.stVar.getComputed(meta)` returns `inline` and `reference` with the same value:

```js
// correct result after fix:
{
  value: { a: 'green' },
  input: {
    type: 'st-map',
    value: {
      a: {
        flatValue: 'green',
        type: 'st-string',
        value: 'green',
    },
  },
}

// previously inline would return the correct result and reference would return:
{
  value: { a: 'green' },
  input: {
    type: 'st-map',
    value: {
      a: 'green',
    },
  },
}
```

### DIFF
--- a/packages/core/src/custom-values.ts
+++ b/packages/core/src/custom-values.ts
@@ -154,6 +154,13 @@ export const CustomValueStrategy = {
                     } else {
                         resolvedValue = unbox(getStringValue(valueNode), !boxPrimitive);
                     }
+                } else if (typeof resolvedValue === 'string') {
+                    const parsedArg = postcssValueParser(resolvedValue).nodes[0];
+                    const ct = parsedArg.type === 'function' && parsedArg.value;
+                    resolvedValue =
+                        typeof ct === 'string' && customTypes[ct]
+                            ? customTypes[ct].evalVarAst(parsedArg, customTypes, boxPrimitive)
+                            : unbox(resolvedValue, !boxPrimitive);
                 }
             } else {
                 resolvedValue = unbox(getStringValue(valueNodes), !boxPrimitive);
--- a/packages/core/src/custom-values.ts
+++ b/packages/core/src/custom-values.ts
@@ -143,27 +143,30 @@ export const CustomValueStrategy = {
             let resolvedValue;
             if (valueNodes.length === 0) {
                 // TODO: error
-            } else if (valueNodes.length === 1) {
-                const valueNode = valueNodes[0];
-                resolvedValue = valueNode.resolvedValue;
-
-                if (!resolvedValue) {
-                    const ct = customTypes[valueNode.value];
-                    if (valueNode.type === 'function' && ct) {
-                        resolvedValue = ct.evalVarAst(valueNode, customTypes, boxPrimitive);
-                    } else {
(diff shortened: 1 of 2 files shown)


packages/core/test/features/st-var.spec.ts
```diff
@@ -1540,11 +1559,29 @@ describe(`features/st-var`, () => {
                     input: {
                         type: 'st-map',
                         value: {
-                            a: 'red',
+                            a: {
```
barak007, line 1562: This is where we actually want a string. 
We can check if the type is "primitive" or check specific for "st-string" and use it's `flatValue`.
When we add types to values (like color in this case) we would print the full type.

One note on `st-string` here. I'm not sure if a value with quotes is `st-string` or just any value. because we might want to change it.

packages/core/src/custom-values.ts
```diff
@@ -154,6 +154,13 @@ export const CustomValueStrategy = {
                     } else {
                         resolvedValue = unbox(getStringValue(valueNode), !boxPrimitive);
                     }
+                } else if (typeof resolvedValue === 'string') {
+                    const parsedArg = postcssValueParser(resolvedValue).nodes[0];
```
barak007, line 158: We want to ignore comments so the `nodes[0]` should be changed to take the first node that is not comment. I assume space cannot be the first token.

packages/core/test/features/st-var.spec.ts
idoros: I think st-string is what you would always want here,  if we had a type for color then the type would change to color.  

In what scenario would you prefer to have a flat value instead? (which you have on the value field already)

barak007: Not sure....it means that all of these are st-string

```
:vars {
    a: "'red'";
    b: "red";
    c: red;
}
```

packages/core/src/custom-values.ts
idoros: the resolved values are already without any comments

idoros: Currently referenced vars are stripped of comments, while the behavior of the top level is strange - it strip any comments before the referenced `value()` and keeps any after... while not related directly to this fix, I added another fix to strip any leftover comments.

packages/core/test/features/st-var.spec.ts
idoros: I'm not sure we need to have a difference in types for these cases. Quotations are used to wrap strings that would otherwise break the syntax. but in the end the result is the same. In any case this PR didn't change the handling of strings, just maps.

barak007 (approved):

PR #2673, opened 2022-08-17
feat(webpack-plugin): expose function to bundle libs according to config

tomrav:
This is to make [using NextJS with Stylable](https://stylable.io/docs/getting-started/nextjs-integration/) easier, and to avoid users having to inline the function in their configuration.

### DIFF
--- a/packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
+++ b/packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
@@ -68,8 +68,8 @@ export function bundleLibs(config: Configuration, packages: string[], isServer: 
                                 return false;
                             }
                         }
-                        return nextExternal(ctx, () => {
-                            // noop
+                        return nextExternal(ctx, (data) => {
+                            (data as any).resolve();
                         });
                     },
                 ];
--- a/packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
+++ b/packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
@@ -55,30 +55,36 @@ function safelyWalkJSON(
     }
 }
 
-export function bundleLibs(config: Configuration, packages: string[], isServer: boolean) {
+export function bundleServerLibs(config: Configuration, packages: string[], isServer: boolean) {
+    let hasError = false;
+
     if (isServer) {
         if (Array.isArray(config.externals) && config.externals.length === 1) {
             const nextExternal = config.externals[0];
 
             if (typeof nextExternal === 'function') {
                 config.externals = [
-                    async (ctx: any) => {
+                    async (ctx: any, cb: any) => {
                         for (const pack of packages) {
                             if (ctx.request.startsWith(pack)) {
                                 return false;
                             }
                         }
-                        return nextExternal(ctx, (data) => {
-                            (data as any).resolve();
-                        });
+                        return nextExternal(ctx, cb);
                     },
                 ];
             } else {
-                throw new Error(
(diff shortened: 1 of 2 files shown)


packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
```diff
@@ -54,3 +54,31 @@ function safelyWalkJSON(
         }
     }
 }
+
+export function bundleLibs(config: Configuration, packages: string[], isServer: boolean) {
+    if (isServer) {
```
idoros, line 59: It's weird to call `bundleLibs` and for some reason it will only work on the server. 
So maybe change the name to `bundleLibsOnServerToPreventNodeImports`?

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
```diff
@@ -54,3 +54,31 @@ function safelyWalkJSON(
...
+        if (Array.isArray(config.externals) && config.externals.length === 1) {
```
idoros, line 60: Is the `externals` field in nextjs always a function config? can something change it before the user has a chance to call our modifier function?

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
```diff
@@ -54,3 +54,31 @@ function safelyWalkJSON(
...
+                                return false;
+                            }
+                        }
+                        return nextExternal(ctx, (data) => {
+                            (data as any).resolve();
+                        });
+                    },
+                ];
+            } else {
+                throw new Error(
```
idoros, line 77: If only called inside the internal else, then it will silently do nothing for the case that the config is an object or a an array with more the a single item.

idoros (changes requested):
No need for a test?

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
```diff
@@ -54,3 +54,31 @@ function safelyWalkJSON(
...
+            if (typeof nextExternal === 'function') {
+                config.externals = [
+                    async (ctx: any) => {
+                        for (const pack of packages) {
+                            if (ctx.request.startsWith(pack)) {
+                                return false;
+                            }
+                        }
+                        return nextExternal(ctx, (data) => {
+                            (data as any).resolve();
```
idoros, line 72: `data.resolve()`? - shouldn't this function be called with `err, result, type?` - https://webpack.js.org/configuration/externals/#function

Why not pass in the 2nd argument callback that the wrapper function takes?

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
tomrav: Changed to `bundleServerLibs`

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
tomrav: I don't think that's the case, but we'll encounter it as it happens.

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
tomrav: fixed

packages/webpack-plugin/src/webpack-config-stylable-excludes.ts
tomrav: fixed

idoros (approved):

PR #2691, opened 2022-08-29
feat(config): add defaultConfig from file

tomrav:
Starting work on resolving https://github.com/wix/stylable-intelligence/issues/930

This PR adds the ability to add a new `defaultConfig` field to the `stylable.config.js` file.
This will be picked up by all of our integrations, and the stylable-intelligence plugin.


packages/rollup-plugin/test/projects/config-file-resolver-override/src/index.st.css
```diff
@@ -0,0 +1,5 @@
+@st-import Comp from "components/comp.st.css";
```
idoros, line 1: green was a better name then comp

packages/rollup-plugin/test/rollup-config-file-resolver.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+describe('StylableRollupPlugin', function () {
+    this.timeout(20000);
+
+    const project = 'config-file-resolver-override';
+
+    const runner = rollupRunner({
+        projectPath: getProjectPath(project),
+        entry: './src/index.js',
+        pluginOptions: {
+            stcConfig: './project/stylable.config.js',
```
idoros, line 16: did you have to provide this? is this because of how the tests are arranged?

packages/rollup-plugin/src/index.ts
```diff
@@ -78,6 +78,21 @@ export function stylableRollupPlugin({
...
             emittedAssets = emittedAssets || new Map();
+
+            const configuration = resolveStcConfig(
+                projectRoot,
+                typeof stcConfig === 'string' ? stcConfig : undefined,
+                fs
+            );
+
+            if (stcConfig && !configuration) {
+                throw new Error(
```
idoros, line 89: does it ever reach this point? wouldn't the internal `require.resolve` throw before reaching this point?

packages/webpack-plugin/test/e2e/stc-config-resolver.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+        const { page } = await projectRunner.openInBrowser();
+        const { backgroundColor, color } = await page.evaluate(() => {
+            const style = getComputedStyle(document.body);
+            return {
+                color: style.color,
+                backgroundColor: style.backgroundColor,
+            };
+        });
+
+        expect(backgroundColor).to.eql('rgb(0, 128, 0)');
```
idoros, line 33: would be nice to have expect labels on fail: `green from webpack alias` and `blue from typescript paths`

packages/webpack-plugin/test/e2e/projects/stylable-config-resolver-override/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+const { StylableWebpackPlugin } = require('@stylable/webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+/** @type {import('webpack').Configuration} */
+module.exports = {
+    mode: 'development',
+    context: __dirname,
+    devtool: 'source-map',
+    plugins: [new StylableWebpackPlugin(), new HtmlWebpackPlugin()],
```
idoros, line 9: missing override/inherit config from the webpack->plugin->config

packages/webpack-plugin/src/plugin.ts
```diff
@@ -294,23 +294,20 @@ export class StylableWebpackPlugin {
...
+    private getStylableConfig(compiler: Compiler) {
         const configuration = resolveStcConfig(
             compiler.context,
-            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined
+            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined,
+            getTopLevelInputFilesystem(compiler)
         );
 
-        if (!configuration) {
+        if (this.options.stcConfig && !configuration) {
```
idoros, line 310: same as rollup - not sure this ever gets here

idoros (changes requested):

packages/cli/src/config/projects-config.ts
```diff
@@ -54,29 +60,32 @@ export async function projectsConfig(
     return projects;
 }
 
-export function resolveConfig(context: string, request?: string) {
-    return request ? requireConfigFile(request, context) : resolveConfigFile(context);
+export function resolveConfig(context: string, request?: string, fs?: MinimalFS) {
```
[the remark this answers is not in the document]
tomrav: I didn't want to break the existing contract for this API (required arg can't follow an optional one)

packages/rollup-plugin/src/index.ts
tomrav: I believe you're right, removing this code.

packages/rollup-plugin/test/projects/config-file-resolver-override/src/index.st.css
tomrav: going back to green

packages/rollup-plugin/test/rollup-config-file-resolver.spec.ts
tomrav: This is because of our setup, I believe for most project structures, passing "true" here will work as intended

packages/webpack-plugin/src/plugin.ts
tomrav: removing as well.

packages/webpack-plugin/test/e2e/stc-config-resolver.spec.ts
tomrav: added

packages/webpack-plugin/test/e2e/projects/stylable-config-resolver-override/webpack.config.js
tomrav: added another test to verify this behavior

packages/jest/test/jest.spec.ts
```diff
@@ -34,4 +36,49 @@ describe('jest process', () => {
...
+        const transformer = stylableTransformer.createTransformer({
+            stylable: {
+                resolveNamespace: (ns, _srcPath) => `${ns}-custom`,
+                resolveModule: createDefaultResolver(fs, {
+                    alias: {
+                        'wp-alias': join(dirname(filename), 'webpack-alias'),
+                    },
+                }),
+            },
+            configPath: join(dirname(filename), 'stylable.config.js'),
```
idoros, line 72: I can't seem to understand the difference between the 2 resolvers, maybe have a single import and let each resolver route to a different file with a different `-st-global` on the same name class

packages/jest/src/jest.ts
```diff
@@ -30,15 +30,28 @@ function getCacheKey(
...
 export interface StylableJestConfig {
     stylable?: Partial<StylableConfig>;
+    configPath?: string;
 }
 
 export const createTransformer = (options?: StylableJestConfig) => {
+    let resolveModule;
+
+    if (options?.configPath && !options.stylable?.resolveModule) {
+        const { defaultConfig } = require(options.configPath);
```
idoros, line 40: would be nice to catch any errors in require or call to user call and through a clear error of where something failed (true for similar places in this PR)

packages/webpack-plugin/src/plugin.ts
```diff
@@ -363,12 +364,12 @@ export class StylableWebpackPlugin {
                     fileSystem: topLevelFs,
                     mode: compiler.options.mode === 'production' ? 'production' : 'development',
                     /**
-                     * resolveModule config order
-                     * 1. webpackPlugin in stylable config file
-                     * 2. defaultConfig in stylable config file
-                     * 3. stylableConfig in webpack config
+                     * resolveModule config order ()
```
idoros, line 367: `()`? - 1 before 3 I assume

packages/node/test/require-hook-default-config.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+describe('require hook', () => {
+    afterEach(() => {
+        delete require.extensions['.css'];
+        for (const itemName of readdirSync(fixturesPath)) {
+            delete require.cache[join(fixturesPath, itemName)];
+        }
+    });
+
+    it('should work on .st.css', () => {
```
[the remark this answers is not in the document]
tomrav: Fixed.

packages/node/test/require-hook-default-config.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+        attachHook({ configPath: join(fixturesPath, 'stylable.config.js') });
+        const indexModule = require(join(fixturesPath, 'index.st.css'));
+        const mappedModule = require(join(fixturesPath, 'webpack-alias/green.st.css'));
+
+        // this checks that a mapped module was correctly resolved using -st-extends
+        expect(indexModule.classes.root).to.equal(
+            indexModule.namespace + '__root' + ' ' + mappedModule.namespace + '__test'
+        );
+    });
+});
```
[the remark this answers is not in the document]
tomrav: Added test.

packages/jest/src/jest.ts
tomrav: Added try/catch and matching test

packages/jest/test/jest.spec.ts
tomrav: refactored this to use two different directories and inline `@st-namespace`

idoros (approved):

packages/webpack-plugin/src/plugin.ts
```diff
@@ -295,40 +295,38 @@ export class StylableWebpackPlugin {
         const options =
             loadStylableConfig(compiler.context, (config) => {
                 return isWebpackConfigProcessor(config)
-                    ? config.webpackPlugin(defaults, compiler)
+                    ? config.webpackPlugin(defaults, compiler, getTopLevelInputFilesystem(compiler))
```
barak007, line 298: why do we need the fs passed here?

packages/webpack-plugin/src/plugin.ts
```diff
@@ -404,7 +419,11 @@ export class StylableWebpackPlugin {
                          * They might be used by other stylesheets so they might end up in the final build
                          */
                         for (const request of stylableBuildMeta.unusedImports) {
-                            module.addDependency(new this.entities.UnusedDependency(request));
+                            module.addDependency(
+                                new this.entities.UnusedDependency(
+                                    this.stylable.resolver.resolvePath(module.context!, request)
```
barak007, line 424: I think its not the place to resolve this. a better place would be where we create `unusedImports`. I will also need to double check this change to make sure we don't somehow use the request field on `UnusedDependency`.

packages/webpack-plugin/src/plugin.ts
```diff
@@ -295,40 +295,38 @@ export class StylableWebpackPlugin {
...
-
+    private getStylableConfig(compiler: Compiler) {
         const configuration = resolveStcConfig(
             compiler.context,
-            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined
+            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined,
+            getTopLevelInputFilesystem(compiler)
         );
 
-        if (!configuration) {
```
barak007, line 314: where did this error go?

packages/webpack-plugin/src/plugin.ts
```diff
@@ -295,40 +295,38 @@ export class StylableWebpackPlugin {
...
-        if (!this.options.stcConfig) {
-            return;
-        }
-
+    private getStylableConfig(compiler: Compiler) {
         const configuration = resolveStcConfig(
             compiler.context,
-            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined
+            typeof this.options.stcConfig === 'string' ? this.options.stcConfig : undefined,
+            getTopLevelInputFilesystem(compiler)
```
barak007, line 308: why we need the fs here?

barak007: I think I know why....because we evaluate the default config inside...I'm not sure we want this...

packages/webpack-plugin/src/plugin.ts
tomrav: To allow the webpack config to use the same FS as webpack itself. 

See the (non-inline) override test for an example.

packages/webpack-plugin/src/plugin.ts
tomrav: `resolveStcConfig` actually has a tryRun internally, so the check here was redundant.

packages/webpack-plugin/src/plugin.ts
tomrav: The reasoning seems right, we can re-think this if we want.

packages/webpack-plugin/src/plugin.ts
idoros: this is changed in #2786

PR #2744, opened 2022-09-29
feat: CSS contains support

idoros:
This PR adds support for CSS containers

- [x] core support
  - support namespaced `container` symbol from `container-name` and `container` declarations
  - `@container` at-rule
  - `st-global` support
  - import between stylesheets
  - mixin support
- [x] JS module
  - `containers` exports
  - Typescript `d.ts` + source maps
  - runtime module
- [x] integrations - webpack & rollup
- [x] language service
- [x] documentation: https://github.com/wixplosives/stylable.io/pull/59


packages/core/src/features/css-contains.ts
```diff
@@ -0,0 +1,345 @@
...
+    let noneFound = false;
+    const checkNextName = (node: valueParser.Node) => {
+        const { type, value } = node;
+        if (type === 'comment' || type === 'space') {
+            // do nothing
+        } else if (type === 'word' || (type === 'function' && value === GLOBAL_FUNC)) {
+            const global = type === 'function';
+            const name = global ? globalValueFromFunctionNode(node) || '' : node.value;
+            if (global && !name) {
+                // report.report(diagnostics.MISSING_LAYER_NAME_INSIDE_GLOBAL(), { node: atRule });
```
tomrav, line 203: Implement commented out diagnostic

tomrav (approved):
Another native feature scoped, hurray! :rocket:

packages/core/src/features/css-contains.ts
idoros: done

packages/module-utils/src/generate-dts-sourcemaps.ts
```diff
@@ -131,6 +131,18 @@ function getLayersSrcPosition(layerName: string, meta: StylableMeta): Position |
 
     return;
 }
+function getContainersSrcPosition(containerName: string, meta: StylableMeta): Position | undefined {
```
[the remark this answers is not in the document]
idoros: will do in a different PR, just noticed there is a probable bug in the same logic for `keyframes`

packages/core/src/features/css-contains.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+    if (containers.length > 0 && noneFound) {
+        context.diagnostics.report(diagnostics.INVALID_CONTAINER_NAME('none'), {
+            node: decl,
+            word: 'none',
+        });
+    }
+    return {
+        containers,
+        transformNames(getTransformedName: (name: string) => string) {
```
barak007, line 280: I thing it might be better to make this function pure and curry the data

packages/core/src/helpers/rule.ts
```diff
@@ -85,7 +85,8 @@ export function createSubsetAst<T extends postcss.Root | postcss.AtRule>(
                 node.name === 'media' ||
                 node.name === 'supports' ||
                 node.name === 'st-scope' ||
-                node.name === 'layer'
+                node.name === 'layer' ||
+                node.name === 'container'
```
barak007, line 89: nice

packages/core/src/stylable-transformer.ts
```diff
@@ -230,6 +239,12 @@ export class StylableTransformer {
                     atRule,
                     resolved: layerResolve,
                 });
+            } else if (name === 'container') {
```
barak007, line 242: I wish that the checks would inline to the hook and see a flat list here, maybe we want to open issue for it.

packages/core/test/features/css-contains.spec.ts
```diff
@@ -0,0 +1,496 @@
...
+        expect(CSSContains.get(meta, 'none'), 'no none symbol').to.eql(undefined);
+
+        // JS exports
+        expect(exports.containers).to.eql({});
+    });
+    it('should mark as global', () => {
+        const { sheets } = testStylableCore(`
+            .a {
+                /* @decl(use global) container-name: a b c d e */
+                container-name: a b c d e;
```
barak007, line 106: so st-global override the local for each symbol?

barak007 (commented):
Looks amazing! 
I don't think I saw usage of css custom properties in a query...but I saw that it is using `"eval"` value.

packages/core/src/stylable-transformer.ts
idoros: I will open an issue the to refactor and unify the code in this area

idoros:
> I don't think I saw usage of css custom properties in a query...but I saw that it is using "eval" value.

The props do go through `evaluateValue` in order handle build vars transformation (see `should transform build vars in @container query` test), but no handling for standalone vars without a `var()` function, since we don't support them yet.

idoros:
**Offline discussion update:** to bring the behavior more inline with custom properties and allow easier and more predictable global definition of container names:

 - Change `container/container-name` decl to soft definition - meaning that it won't cause a redeclare diagnostic and override imported names with local ones.
 - `st-global` should mark a global name in both `@container` atrule and `container/container-name` without defining a symbol
 - new `@container name;` atrule with no body for hard definition + st-global to create a global symbol

packages/core/test/features/css-contains.spec.ts
idoros: resolved in offline discussion: https://github.com/wix/stylable/pull/2744#issuecomment-1310035240

packages/core/test/features/css-contains.spec.ts
```diff
@@ -369,12 +403,54 @@ describe('features/css-contains', () => {
                 `,
             });
         });
-        it('should resolve imported global container name', () => {
+        it('should report redeclare of imported container name', () => {
             const { sheets } = testStylableCore({
-                '/imported.st.css': `
+                '/a.st.css': `
```
barak007, line 408: check double declaration in single file with and without `st-global`

idoros: good catch!

barak007 (approved):

PR #2746, opened 2022-10-03
feat: support native CSS file import

idoros:
This PR adds the ability to import and use references from a native CSS stylesheet in a Stylable stylesheet.

- [x] analyze `.css` file into a stylable meta
  - collect global symbols for `classes`, `custom properties`, `keyframes`, and `layers`
  - ignore stylable directives
  - prevent stylable specific transformations (`st-global` and some code removal)
- [ ] bundle `.css` files in webpack
- [ ] bundle `.css` files in rollup
- [ ] copy relative `.css` files as assets when copying `.st.css` files


packages/cli/src/build-single-file.ts
```diff
@@ -213,6 +213,12 @@ export function buildSingleFile({
             projectAssets.add(resolve(fileDirectory, url));
         }
     }
+    // add native css imports as assets
+    for (const { request } of res.meta.getImportStatements()) {
+        if (request.endsWith('.css') && !request.endsWith('.st.css') && isAsset(request)) {
```
barak007, line 218: It's problematic to use the request to determine if a file is Stylable. we should first resolve the request and check on that

packages/rollup-plugin/src/index.ts
```diff
@@ -55,6 +55,19 @@ const clearRequireCache = () => {
 };
 
 const ST_CSS = '.st.css';
+const LOADABLE_CSS_QUERY = '?stylable';
```
barak007, line 58: maybe change to ?stylable-plain-css

packages/rollup-plugin/src/index.ts
```diff
@@ -210,26 +229,27 @@ export function stylableRollupPlugin({
             };
 
             for (const moduleId of this.getModuleIds()) {
-                if (moduleId.endsWith(ST_CSS)) {
-                    modules.push({ depth: calcDepth(moduleId, context, [], cache), moduleId });
+                const { isStFile, isLoadableCssFile, path } = getLoadableModuleData(moduleId);
+                if (isStFile || isLoadableCssFile) {
+                    modules.push({ depth: calcDepth(moduleId, context, [], cache), path });
```
barak007, line 234: why do we calc depth for the css? I'm not sure that algorithm knows how to handle it.

packages/rollup-plugin/src/index.ts
```diff
@@ -210,26 +229,27 @@ export function stylableRollupPlugin({
...
                 }
             }
 
             sortModulesByDepth(
                 modules,
                 (m) => m.depth,
-                (m) => m.moduleId,
+                (m) => m.path,
```
barak007, line 241: why change?

packages/rollup-plugin/src/index.ts
```diff
@@ -210,26 +229,27 @@ export function stylableRollupPlugin({
...
                 PRINT_ORDER
             );
 
             outputCSS = '';
 
-            for (const { moduleId } of modules) {
-                const stored = extracted.get(moduleId);
+            for (const { path } of modules) {
```
barak007, line 247: I'm not sure all modules have path...

packages/webpack-plugin/src/loader-utils.ts
```diff
@@ -17,9 +18,13 @@ export function getImports(
     const imports: string[] = [];
     const unusedImports: string[] = [];
     for (const imported of meta.getImportStatements()) {
-        if (imported.request.endsWith('.st.css')) {
+        if (imported.request.endsWith('.css')) {
```
barak007, line 21: this was wrong in the first place. we should not check the request. we should check the resolved value.

barak007 (commented):
what about the experimental loader? do we want to update it?

packages/rollup-plugin/src/index.ts
idoros: it doesn't mean anything. will always be 0

packages/rollup-plugin/src/index.ts
idoros: its just the inline type from a few lines back changing. there is a different between `moduleId` and `path` now that we add the query param to native CSS modules

packages/rollup-plugin/src/index.ts
idoros: they do because we create the module list in this scope and the path is created from the module id minus any query params

packages/webpack-plugin/test/e2e/native-css.ts
```diff
@@ -27,26 +27,42 @@ describe(`(${project})`, () => {
...
-                const computedStyle = getComputedStyle((window as any).document.body);
-                return {
-                    localSideEffect: computedStyle.getPropertyValue('--local-side-effect'),
-                    libSideEffect: computedStyle.getPropertyValue('--lib-side-effect'),
-                    libClass: computedStyle.getPropertyValue('--lib-class'),
-                    localColor: computedStyle.backgroundColor,
-                    libColor: computedStyle.color,
-                };
-            });
+        const {
```
barak007, line 30: add .spec to the file

packages/core/src/features/st-import.ts
```diff
@@ -97,6 +97,11 @@ export const diagnostics = {
         'error',
         (type: string) => `Unknown type import "${type}"`
     ),
+    NO_DEFAULT_EXPORT: createDiagnosticReporter(
+        '05020',
+        'error',
+        (path: string) => `No default export for imported file: "${path}"`
```
tomrav, line 103:
```suggestion
        (path: string) => `Native CSS files have no default export. Imported file: "${path}"`
```

tomrav (approved):
Should add a check that native CSS files do not include @import statements to other stylesheets.

Great work!

tomrav (approved):

packages/core/src/stylable-assets.ts
```diff
@@ -23,6 +23,15 @@ export function isAsset(url: string) {
     return !isExternal(url);
 }
 
+export function isRelativeNativeCss(request: string) {
```
barak007, line 26: this is not a request here...it's a `fullpath`

barak007 (approved):

PR #2783, opened 2022-11-23
fix: webpack optimization bailout for generate stylable modules

barak007:
This PR uses `HarmonyImportDependency` (exposed here [webpack PR](https://github.com/webpack/webpack/pull/16032)), to describe unused dependencies connections, this change causes webpack to concat the resulting code for all the resulting JS modules created by stylesheets, and make inline optimization to work.


idoros (approved):
Seems ok, it's really hard to know what this might effect since its another part of webpack with almost no documentation. Lets merge and fix if something comes up 🤷

PR #2786, opened 2022-11-29
feat: improve side effect detection

idoros:
This PR should improve detection of stylesheets with side-effects to be auto loaded without any extra imports in JavaScript files. 

Currently a stylesheet is transformed only with imports that are considered to have direct side-effects on it:
- direct import of `keyframes` or `layer`
- direct import of a composed class

Cases that are missing and should improve the developer experience:
- direct import check
  - [ ] add `property` to the existing direct import side effect check
- deep imports to stylesheets that contain side-effects (globals)
  - [ ] at-rules (keyframes, layer, property)
  - [ ] rules with global selectors (with no local/namespaced part)
  - [ ] need to research: mixins that are (1) applied on global rules and generates global rules, and (2) inject global at-rules.


packages/webpack-plugin/test/e2e/side-effects.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        afterEach,
+        after
+    );
+
+    it('should include deep sheets with side effects', async () => {
+        const { page } = await projectRunner.openInBrowser();
+        const styleElements = await page.evaluate(browserFunctions.getStyleElementsMetadata);
+
+        expect(styleElements).to.eql([
+            { id: './src/native.css', depth: '0' },
```
barak007, line 31: note here about the zero

packages/webpack-plugin/src/loader-utils.ts
```diff
@@ -1,46 +1,86 @@
...
-            resolved = stylable.resolver.resolvePath(imported.context, imported.request);
-        } catch (e) {
-            // fallback to request
+    if (includeGlobalSideEffects) {
+        // new mode that collect deep side effects
+        const resolvedSideEffectRequests = collectSheetsWithSideEffects(stylable, meta);
+        const contextDir = dirname(meta.source);
+        // add side-effects sheets as imports
+        for (const resolvedRequest of resolvedSideEffectRequests) {
+            const relativePath = normalizeRelative(relative(contextDir, resolvedRequest));
```
barak007, line 38: webpack and rollup don't need relative paths in the modules. make sure to remove the resolve in plugin.ts
for both modes!

barak007: in the Cli it will be important to use the relative

barak007 (changes requested):

packages/core/src/features/st-global.ts
```diff
@@ -27,21 +31,62 @@ export const diagnostics = {
...
             });
         }
-        return walkSelector.skipNested;
+        const ruleData = rules.get(rule)!;
+        if (node.type === 'pseudo_class' && node.value === `global`) {
+            // mark selector as global only if it isn't set
+            if (ruleData.selectors[topSelectorIndex] === undefined) {
+                ruleData.selectors[topSelectorIndex] = true;
+            }
+            if (node.nodes && node.nodes?.length > 1) {
```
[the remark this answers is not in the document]
idoros: was already like this, but changing this has 2 issues:
- typescript doesn't like the undefined compare: "Object is possibly 'undefined'.ts(2532)"
- typescript doesn't infer the nodes as existing - forcing a `!` when passing them in the nested code

barak007 (approved):

PR #2787, opened 2022-11-30
feat: expand resolution error with error details

barak007:
propagate the resolve module error to the diagnostic.

### DIFF
--- a/packages/core/src/features/st-import.ts
+++ b/packages/core/src/features/st-import.ts
@@ -10,6 +10,7 @@ import type { ImmutablePseudoClass, PseudoClass } from '@tokey/css-selector-pars
 import type * as postcss from 'postcss';
 import { createDiagnosticReporter } from '../diagnostics';
 import type { Stylable } from '../stylable';
+import type { CachedModuleEntity } from '../stylable-resolver';
 
 export interface ImportSymbol {
     _kind: 'import';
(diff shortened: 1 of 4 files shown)


packages/core/src/features/st-import.ts
```diff
@@ -275,10 +276,16 @@ function validateImports(context: FeatureTransformContext) {
...
                 );
 
-            context.diagnostics.report(diagnostics.UNKNOWN_IMPORTED_FILE(importObj.request), {
-                node: fromDecl || importObj.rule,
-                word: importObj.request,
-            });
+            context.diagnostics.report(
+                diagnostics.UNKNOWN_IMPORTED_FILE(
+                    importObj.request,
+                    context.resolver.getErrorText(importObj)
```
idoros, line 282: wouldn't it be better to call the module once? If you don't want to change `resolveImported`, then maybe add a new method to the resolver that returns a `[result, error]` ?

It would be nice to have something like `getModule` public without in order to get a JS module or stylable meta from import without having to pass symbol name.

packages/core/src/features/st-import.ts
```diff
@@ -90,7 +90,8 @@ export const diagnostics = {
     UNKNOWN_IMPORTED_FILE: createDiagnosticReporter(
         '05016',
         'error',
-        (path: string) => `cannot resolve imported file: "${path}"`
+        (path: string, error?: unknown) =>
```
idoros, line 93: why is the error optional? it's only called in one place that has an error to pass

idoros (changes requested):

packages/core/src/features/st-import.ts
barak007: If make `getModule` public we could. from what I hear that's what you want?

barak007: it will improve the surrounding of this function that uses `""` for the symbol name

packages/core/src/features/st-import.ts
barak007: there are tests that does not concern that and not passing error here...they check with `include`

idoros: ok

packages/core/src/features/st-import.ts
idoros: is there an issue with making it public?

idoros (approved):

PR #2798, opened 2022-12-19
feat: remove stylable directives at transform

idoros:
This PR moves the removal of `-st-*` declarations to be part of the transformation when used with `mode: production`. 

In addition the `removeStylableDirectives` configuration is removed and a `console.warn` with an appropriate deprecation/unrequired message is logged.


tomrav (approved):
Looks good!

PR #2799, opened 2022-12-21
chore: replace deindent dependency

idoros:
This PR removed the dependency on `deindent` package. It has several issues, like losing newlines and in some cases spaces, and in addition doesn't trim the first/last empty lines, making it harder to use.

- remove deindent as `@stylable/core` dependency - not sure why it wasn't a dev-dependency as it's only used in tests
- write alternative into `@stylable/core-test-kit`
    - trim empty first/last lines
    - remove indentation based on the minimal length of indentation before content
    - preserve newlines and spaces
- replace all usages - had to add `core-test-kit` as a dependency to several test folders

### DIFF
--- a/packages/code-formatter/test/formatter.spec.ts
+++ b/packages/code-formatter/test/formatter.spec.ts
@@ -1,6 +1,6 @@
+import { deindent } from '@stylable/core-test-kit';
 import { getDocumentFormatting } from '@stylable/code-formatter';
 import { expect } from 'chai';
-import deindent from 'deindent';
 
 describe('Formatting', () => {
     it('should format an entire stylesheet with extra spaces', () => {
@@ -30,7 +30,8 @@ describe('Formatting', () => {
         );
 
         expect(res).to.eql(
-            deindent(`/* PRESERVE INDENT */
+            deindent(`
+                /* PRESERVE INDENT */
                 .foo {
                     grid-template:
                         "icon  name name ."    16px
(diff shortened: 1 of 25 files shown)


packages/code-formatter/test/formatter.spec.ts
```diff
@@ -30,7 +30,8 @@ describe('Formatting', () => {
         );
 
         expect(res).to.eql(
-            deindent(`/* PRESERVE INDENT */
+            deindent(`
+                /* PRESERVE INDENT */
```
tomrav, line 34: Can remove.

packages/core-test-kit/src/deindent.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Naive deindent - takes in a string and:
+ *  1. find the minimal indentation across all lines with content
```
tomrav, line 3:
```suggestion
 *  1. finds the minimal indentation across all lines of content
```

packages/core-test-kit/src/deindent.ts
```diff
@@ -0,0 +1,29 @@
...
+ *  2. removes the minimal whitespace for each line
+ *  3. attempt to remove first and last line in case of empty lines to improve usage
```
tomrav, line 5:
```suggestion
 *  3. attempts to remove first and last line in case of empty lines to improve usage
```

tomrav (approved):
Left tiny fixes, looks great - farewell `deindent`, hello `deindent`.

PR #2802, opened 2023-01-06
Fix `@stylable/core/src/types` import in public API

amitdahan:
# Summary

This PR fixes a bad import pointing to `@stylable/core/src/types` that causes consumers who run `tsc` to also type-check `@stylable/core`'s `.ts` sources even if they have `skipLibCheck` configured (due to having a direct import-path to it).

I added a `@stylable/core/types` export to point to the correct, transpiled file.


idoros:
Thank you for the help. The import from `src` is definitely an oversight, but can you help me understand how to reproduce the typecheck error?

amitdahan:
Of course, all I did was (`typescript@^4.9.4`) run `tsc --noEmit` with the following:

- `tsconfig.json`:
```json
{
  "compilerOptions": {
    "strict": true,
    "skipLibCheck": true,
    "moduleResolution": "node"
  },
  "files": ["index.ts"]
}

```

- `index.ts`:
```ts
import { StylableWebpackPlugin } from '@stylable/webpack-plugin';

console.log(StylableWebpackPlugin);
```

packages/cli/src/config/projects-config.ts
```diff
@@ -13,7 +13,7 @@ import type {
 import { processProjects } from './process-projects';
 import { createDefaultOptions, mergeBuildOptions, validateOptions } from './resolve-options';
 import { resolveNpmRequests } from './resolve-requests';
-import type { ModuleResolver } from '@stylable/core/src/types';
+import type { ModuleResolver } from '@stylable/core/types';
```
idoros, line 16:
```suggestion
import type { ModuleResolver } from '@stylable/core/dist/index-internal';
```

you can add the `ModuleResolver` to the existing export in core index-internal

packages/core/types.d.ts
```diff
@@ -0,0 +1 @@
+export * from './dist/types';
```
idoros, line 1: I prefer to generate our types during build

idoros (changes requested):
While this fix the issue, we use an internal index to share definitions across packages using `@stylable/core/dist/index-internal` (until we move to project exports).

I left related comments, feel free to fix and I will merge, or I can do it if you prefer, and we will publish a patch version.

amitdahan:
Thanks for the review!

Lmk if this is what you meant:

https://github.com/wix/stylable/pull/2802/commits/79a226450827c8e7e6908490a65fc2627bd97d34

idoros (approved):

idoros:
@amitdahan - I released [5.6.1](https://github.com/wix/stylable/releases/tag/v5.6.1) - thank you for the work and please tell me if the problem persists

barak007:
@amitdahan nice catch.

tzachbon:
@idoros @barak007 - This rule might help for such cases:
https://eslint.org/docs/latest/rules/no-restricted-imports

```
"no-restricted-imports": ["error", {
    "patterns": ["@stylable/*/src/*"]
}]
```

PR #2805, opened 2023-01-11
refactor: simpler mocha entrypoint resolution

AviVahl:
- simplified launch.json: remove useless "Launch Program" and use "type": "node" (as recommended by vscode these days).
- fixed a type error about a variable used before assignment in test runner.


idoros (approved):
I wanted to remove this "launch" configuration for so long and never did...

Issue #2773, opened 2022-10-29
Stylesheet does not get applied when using Next.js 13


srmagura:
### Actual Behavior

Repro project: https://github.com/srmagura/stylable-testbed

In dev and prod, the styles are not applied. In production mode, the stylable CSS is added as a `<script>` tag, which may provide a clue as to what is happening.

**Edit:** Actually, I think this is a problem with the new `app` directory rather than Next.js 13 itself. I created a different project which I thought was Next.js 12, but I was actually using Next.js 13 with `pages`. It works in this second project.

### Expected Behavior

The Stylable styles are applied to the document.

### Version

@stylable/core@5.2.1, @stylable/webpack-plugin@5.2.1

### Relevant log output

```shell
Not getting anything interesting in the logs.
```

tomrav:
Hey, we just released a new version of Stylable (`v5.7.0`), which includes a new webpack flag, meant for next 13 app directory mode.

See more about how to use it here: https://stylable.io/docs/getting-started/nextjs-integration/#app-directory

We'd love to hear back if it worked as expected, and if not what went wrong.

PR #2808, opened 2023-01-15
feat(webpack): experimental flag for next 13 app directory

idoros:
This PR adds an experimental flag `experimentalAttachCssToContainingChunks` that changes the way the CSS asset is attached to webpack chunks - in the default (off mode), the result generated CSS asset is attached to entrypoint chunks, and when the flag is enabled, it attaches the CSS to every chunk that contains stylable stylesheet.  

This is primarily added to test a fix for Next 13 new experimental [app directory feature](https://beta.nextjs.org/docs/routing/fundamentals#the-app-directory) (raised in #2773). In the default mode Next resolve the attached generated CSS module mistakenly as a JavaScript file and attach it in the HTML as a script tag. 

**Notice 1:** I'm adding this experimental flag that seems to resolve the issue, although unfortunately nor NextJS or Webpack document how chunks are used in a clear way or not at all. So this will probably need more investigations and fixes in the future.    

**Notice 2:** There are no tests for this in this PR as there is no NextJS integration in the Stylable repo - we might want to add a separate repository to test such integrations without adding more dependencies.

also added Stylable.io documentation about this configuration: https://github.com/wixplosives/stylable.io/pull/64

### DIFF
--- a/packages/webpack-plugin/src/plugin.ts
+++ b/packages/webpack-plugin/src/plugin.ts
@@ -9,7 +9,7 @@ import { createNamespaceStrategyNode } from '@stylable/node';
 import { sortModulesByDepth, loadStylableConfig, calcDepth } from '@stylable/build-tools';
 import { StylableOptimizer } from '@stylable/optimizer';
 import cloneDeep from 'lodash.clonedeep';
-import type { Compilation, Compiler, NormalModule, WebpackError } from 'webpack';
+import type { Compilation, Compiler, Module, NormalModule, WebpackError } from 'webpack';
 
 import {
     getStaticPublicPath,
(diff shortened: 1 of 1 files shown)


barak007 (approved):

barak007:
We want to write docs for that in next integration page

idoros:
Added a PR to stylable.io: https://github.com/wixplosives/stylable.io/pull/64

barak007 (approved):

barak007 (approved):

packages/webpack-plugin/src/plugin.ts
```diff
@@ -142,6 +142,11 @@ export interface StylableWebpackPluginOptions {
      * Defaults to true.
      */
     includeGlobalSideEffects?: boolean;
+    /**
+     * Experimental flag that attaches CSS bundle asset to every chunk that contains references to stylable stylesheets.
+     * The default off mode attaches the only to entry chunks.
```
tomrav, line 147:
```suggestion
     * The default off mode attaches only to entry chunks.
```

tomrav (approved):
Looks good!

PR #2819, opened 2023-02-03
feat: css-media

barak007:
refactor: use only one walk in transform
refactor: move valueHook to evaluator

### DIFF
--- a/packages/core/src/functions.ts
+++ b/packages/core/src/functions.ts
@@ -22,7 +22,6 @@ export interface EvalValueData {
     value: string;
     passedThrough: string[];
     node?: postcss.Node;
-    // valueHook?: replaceValueHook;
     meta: StylableMeta;
     stVarOverride?: Record<string, string> | null;
     cssVarsMapping?: Record<string, string>;
--- a/packages/core/src/functions.ts
+++ b/packages/core/src/functions.ts
@@ -22,7 +22,7 @@ export interface EvalValueData {
     value: string;
     passedThrough: string[];
     node?: postcss.Node;
-    valueHook?: replaceValueHook;
+    // valueHook?: replaceValueHook;
     meta: StylableMeta;
     stVarOverride?: Record<string, string> | null;
     cssVarsMapping?: Record<string, string>;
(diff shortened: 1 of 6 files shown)


packages/core/src/functions.ts
```diff
@@ -40,21 +39,29 @@ export interface EvalValueResult {
...
+        getResolvedSymbols: (meta: StylableMeta) => MetaResolvedSymbols;
+    }) {
+        this.valueHook = options.valueHook;
         this.stVarOverride = options.stVarOverride;
+        this.getResolvedSymbols = options.getResolvedSymbols;
     }
     evaluateValue(
-        context: FeatureTransformContext,
-        data: Omit<EvalValueData, 'passedThrough'> & { passedThrough?: string[] }
+        context: Omit<FeatureTransformContext, 'getResolvedSymbols'>,
```
idoros, line 54: you just added `passedThrough` to the transform context, and it also sounds like it belongs there more then to data.

Maybe pass is as part of the context and not the data?

packages/core/src/stylable-transformer.ts
```diff
@@ -178,44 +183,39 @@ export class StylableTransformer {
...
 
-        ast.walkRules((rule) => {
-            if (isChildOfAtRule(rule, 'keyframes')) {
-                return;
-            }
-            rule.selector = this.scopeRule(meta, rule, topNestClassName);
-        });
-
-        ast.walkAtRules((atRule) => {
+        const handleAtRule = (atRule: postcss.AtRule) => {
```
idoros, line 211: would be nice to extract this functions

idoros (approved):
looks good

packages/core/src/stylable-transformer.ts
barak007: I tried but it yields bigger change.

packages/core/src/functions.ts
idoros: I added this in the last commit, and added a related issue I fount that wan't caused by any change in this PR: https://github.com/wix/stylable/issues/2820

PR #2831, opened 2023-03-06
refactor: cache selector transform origin

idoros:
When transforming a `pseudo-element/class/type` selector, Stylable is looking for the stylesheet+symbol that defines how the selector should be transformed. This PR caches the lookup part for a single transformation.

This is not a huge performance win for what Stylable does today, but will help with the work on selector intersection types that requires a lot of comparisons that use this lookup.


idoros:
While we want to cache this process, this seems like the wrong direction, its messy and the place that needs it the most (selector intersection) doesn't need all of the stylesheet symbols while looking at a single class/element-type.

Closing this and will probably add similar caching at the transformer level at some other time.

PR #2839, opened 2023-03-21
refactor: make core compatible with esm bundling

AviVahl:
this ensures core can be bundled using esbuild to cjs/esm


idoros:
Should we add some bundle test to ensure changes in the future preserve compatibility?

PR #2842, opened 2023-03-28
feat: css nesting

idoros:
This PR adds support for [native CSS nesting](https://drafts.csswg.org/css-nesting/).

Most of the work to support nesting was already done in the refactor to support [selector intersection](#2827 ).

## Tasks

- [x] remove nesting rules error diagnostic
- [x] test nesting selector (`&`) transformation
- [x] test nesting selector completions in language service
- [x] handle unscoped nested selector - selector not starting with `&` should infer to a universal selector and not resolve to the stylesheet root.

### DIFF
--- a/packages/core/test/stylable-transformer/nesting.spec.ts
+++ b/packages/core/test/stylable-transformer/nesting.spec.ts
@@ -1,0 +1,94 @@
+import {
+    diagnosticBankReportToStrings,
+    shouldReportNoDiagnostics,
+    testStylableCore,
+} from '@stylable/core-test-kit';
+import { CSSPseudoClass } from '@stylable/core/dist/features';
+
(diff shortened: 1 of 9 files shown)


packages/core/test/stylable-transformer/nesting.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+                */
+                :x {}
+
+                /* 
+                    @transform-error(after combinator) ${cssPseudoClassDiagnostics.UNKNOWN_STATE_USAGE(
+                        'x'
+                    )}
+                    @rule(after combinator) .entry__root :x 
+                */
+                .root :x {}
```
barak007, line 116: add the case at the root level to indicate what we do now!

barak007 (approved):

PR #2844, opened 2023-03-29
feat: default to universal inferred context selector

idoros:
This PR changes the inferred selector context under the new `experimentalSelectorInference` flag.

The default inferred selector context is used in 2 cases
- the starting point for any selector 
- reset after a combinator (with `experimentalSelectorInference`)

Before this change the inferred selector used was the stylesheet root, and this PR changes it to be a universal selector (behind the flag), which means that custom-pseudo-element/class cannot be floating:

```css
.root {
    -st-states: state;
}
.part {}

/* all of these cases will now report unknown pseudo diagnostics and will not transform */
:state {}
::part {}
.x :state {}
.x ::part {}
```

### DIFF
--- a/packages/core/test/stylable-transformer/general.spec.ts
+++ b/packages/core/test/stylable-transformer/general.spec.ts
@@ -77,8 +77,22 @@ describe('Stylable postcss transform (General)', () => {
     });
 
     describe('experimentalSelectorInference', () => {
+        it('should set default inferred selector context to universal selector', () => {
+            testStylableCore(
+                `
+                    .root { -st-states: state; }
+                    .class { -st-states: state; }
+                
+                    /* @rule(unknown state) :state */
+                    :state {}
+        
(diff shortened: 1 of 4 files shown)


packages/core/test/stylable-transformer/general.spec.ts
```diff
@@ -77,8 +77,22 @@ describe('Stylable postcss transform (General)', () => {
     });
 
     describe('experimentalSelectorInference', () => {
+        it('should set default inferred selector context to universal selector', () => {
+            testStylableCore(
+                `
+                    .root { -st-states: state; }
+                    .class { -st-states: state; }
+                
+                    /* @rule(unknown state) :state */
```
barak007, line 86: There errors emitted here let's test them. I don't see any other place that test them.

barak007 (approved):

PR #2855, opened 2023-04-23
feat: support CSS nesting in CSS mixin

idoros:
## Custom marker from mixin anchor

Up until now, when a mixin fragment was created, Stylable used a nesting selector (`&`) internally to mark the mixin anchor that would be fused in transformation to the mixin target. This is conflicting with native CSS nesting in case the mixin has a nesting selector in it's source.

This PR refactors the internal selector that is used to mark the mixin anchor from `&` to `[st-mixin-marker]`, solving an issue with mixins that contain CSS nesting.

## Collect and merge nested rules

- Merges the transformed mixin fragment for nested rules
- fix: merge nested at-rules instead of flattening deep declarations

## NO NESTING IN JS MIXINS

JS mixin uses `&` to reference the mix target anchor, which is a public API that cannot change. To support nesting in JS mixin, Stylable would have to offer a new mixin API that would provide some other way of referencing the mixin anchor.

### DIFF
--- a/packages/core/src/helpers/rule.ts
+++ b/packages/core/src/helpers/rule.ts
@@ -146,14 +146,13 @@ export function createSubsetAst<T extends postcss.Root | postcss.AtRule | postcs
 
 export const stMixinMarker = 'st-mixin-marker';
 export const isStMixinMarker = (node: SelectorNode) =>
-    node.type === 'pseudo_class' && node.value === stMixinMarker;
+    node.type === 'attribute' && node.value === stMixinMarker;
 function replaceTargetWithMixinAnchor(selectorNode: Selector, prefixType: ImmutableSelectorNode) {
     walkSelector(selectorNode, (node) => {
         if (matchTypeAndValue(node, prefixType)) {
             convertToSelector(node).nodes = [
                 {
-                    type: `pseudo_class`,
-                    colonComments: [],
+                    type: `attribute`,
                     value: stMixinMarker,
                     start: node.start,
                     end: node.end,
--- a/packages/core/src/helpers/rule.ts
+++ b/packages/core/src/helpers/rule.ts
@@ -48,7 +48,7 @@ export function createSubsetAst<T extends postcss.Root | postcss.AtRule | postcs
     const prefixType = prefixSelectorList[0].nodes[0];
     const containsPrefix = containsMatchInFirstChunk.bind(null, prefixType);
     const mixinRoot = mixinTarget ? mixinTarget : postcss.root();
-    [...root.nodes].forEach((node) => {
+    root.nodes.forEach((node) => {
         if (node.type === 'decl') {
             mixinTarget?.append(node.clone());
         } else if (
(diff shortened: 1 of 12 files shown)


packages/core/src/helpers/selector.ts
```diff
@@ -203,7 +204,7 @@ export function scopeNestedSelector(
                 : false;
             let nestedMixRoot = false;
             walkSelector(outputAst, (node, i, nodes) => {
-                if (node.type === 'nesting') {
+                if (isAnchor ? isAnchor(node) : node.type === 'nesting') {
```
barak007, line 207: why the check of nesting still here?

packages/core/test/helpers/rule.spec.ts
```diff
@@ -54,26 +54,26 @@ describe(`helpers/rule`, () => {
...
-                { selector: '&:not(.x)' },
-                { selector: '& &.x:hover' },
-                { selector: '&.y.x' },
-                { selector: '&&' }, // TODO: check if possible
-                { selector: ':not(&) &' },
-                { selector: ':nth-child(5n - 1 of &)' },
-                { selector: ':nth-child(5n - 2 of &, &)' },
-                { selector: ':nth-child(5n - 3 of &, .x, &)' }, // ToDo: check if to remove unrelated nested selectors
-                { selector: '&' },
+                { selector: ':st-mixin-marker .x' },
```
barak007, line 57: we need to talk about elements and this marker

packages/core/src/stylable.ts
```diff
@@ -185,7 +185,7 @@ export class Stylable {
     ): { selector: string; resolved: ResolvedElement[][] } {
         const meta = typeof pathOrMeta === `string` ? this.analyze(pathOrMeta) : pathOrMeta;
         const transformer = this.createTransformer(options);
-        const r = transformer.scopeSelector(meta, selector, undefined, undefined, true);
+        const r = transformer.scopeSelector(meta, selector, undefined, undefined, undefined, true);
```
barak007, line 188: is this correct?

packages/core/src/helpers/rule.ts
```diff
@@ -47,8 +48,13 @@ export function createSubsetAst<T extends postcss.Root | postcss.AtRule>(
     const prefixType = prefixSelectorList[0].nodes[0];
     const containsPrefix = containsMatchInFirstChunk.bind(null, prefixType);
     const mixinRoot = mixinTarget ? mixinTarget : postcss.root();
-    root.nodes.forEach((node) => {
-        if (node.type === `rule` && (node.selector === ':vars' || node.selector === ':import')) {
+    [...root.nodes].forEach((node) => {
```
barak007, line 51: why do we clone the here if you clean each append?

packages/core/src/helpers/selector.ts
idoros: because for JS mixins we still use &. This will be the case until we deprecate v1 of mixins... first need to plan for v2...
I moved the default to the util args to make it easier to read

packages/core/src/stylable.ts
idoros: yes it is. arg #6 is stripGlobals

packages/core/src/helpers/rule.ts
idoros: removed

barak007 (approved):

PR #2857, opened 2023-05-01
feat: support typed named imports completion

idoros:
This PR refactors the language-service `@st-import` named completions implementation to use the new context, adds supports for current  typed named imports: `keyframes()/layer()/container()`, and fixes up some completion behavior.

- [x] feat: support typed imports
- [x] feat: add completions for typed asserts (e.g. `keyframes($0)`)
- [x] feat: informative detail data for each type of import
    - some relative path cases are not working because `deepResolve` doesn't handle `keyframes`
- [x] fix: relative path and name of origin import
- [x] refactor completion implementation to use new context
- [x] refactor tests to new test-kit

> This PR does not add or change the functionality of `-st-named` from the legacy `:import {}` rule


packages/language-service/src/lib-new/features/ls-st-import.ts
```diff
@@ -50,10 +56,275 @@ function getStImportCompletions(context: LangServiceContext, _importNode: postcs
...
+            deltaStart = -nameBeforeCaret.length;
+        } else {
+            continue;
+        }
+        const originSymbolOrValue = resolveOrigin?.(value);
+        let relativePath = path.relative(
+            path.dirname(context.meta.source),
+            originSymbolOrValue?.originPath || ''
+        );
+        if (!path.isAbsolute(relativePath) && !relativePath.match(/^\./)) {
```
barak007, line 171: .startsWith('.')

packages/language-service/src/lib-new/features/ls-st-import.ts
```diff
@@ -50,10 +56,275 @@ function getStImportCompletions(context: LangServiceContext, _importNode: postcs
...
+            createCompletion({
+                ...originSymbolOrValue,
+                localName: name,
+                rng: range(context.getPosition(), { deltaStart }),
+                relativePath,
+            })
+        );
+    }
+}
+function getSpecifierModule(context: LangServiceContext) {
```
barak007, line 185: we have import parser. but since we already parsed it maybe it's fine?

barak007: looking deeper seems like you need positions so the import parser is not good here

packages/language-service/src/lib-new/features/ls-st-import.ts
idoros: yes... I started with a more complex check and didn't notice I don't need the regex

packages/language-service/test/lib-new/features/ls-st-import.spec.ts
```diff
@@ -13,44 +15,730 @@ describe('LS: st-import', () => {
...
-                'native.css': `
-                    .classA {}
-                    .classB {
-                        --propA: 1;
+                'a.st.css': `
+                    ^topLevel^
+                `,
+                'b.st.css': `
+                    /* existing in file */
+                    @st-import from "./a.st.css";
```
barak007, line 28: remove from.

barak007 (approved):

Issue #2848, opened 2023-04-04
proposal: unify selector definition API


idoros:
# WIP

This proposal tries to unify the way selector API is defined in stylable. It's replacing previous [st-part](#2713 ) proposal for a broader solution aims to make the syntax simpler while adding capabilities that have been long requested and required for future experience we talked about. 

## Goals

- Allow deep pseudo-element definition under the same stylesheet
- Allow multiple components in a single stylesheet
- A way forward to define native CSS/DOM APIs (e.g. custom elements and attributes)
- Unified syntax

## Base Proposal

Single at-rule directive to enrich CSS definitions.

### Define a Component

```css
/* mark the .btn class as a component */
@st .btn;
/* or more verbosely */
@st-comp .btn;
```

### Define Pseudo States

Nested pseudo-states are defined as nested `@st` following a pseudo-class syntax _(colon+name)_. 
To set a non boolean state the name can be followed with parentheses with type definition similar to how [states are defined today](https://stylable.io/docs/references/pseudo-classes#enum).

```css
@st .btn {
    /* boolean state */
    @st :toggled;
    /* enum state */
    @st :size(enum(small, medium, large));
}
```

**Equivalent today**
```css
/* btn.st.css */
.root {
    -st-states: 
        toggled, 
        size(enum(small, medium, large));
}
```
**usage**
```css
/* selector api usage */
.btn:toggled {}
.btn:size(small) {}
```

### Define Pseudo Elements

Nested pseudo-elements are defined as nested `@st` following a pseudo-element syntax _(double colon+name)_ with fat-arrow mapping them to a selector a relative selector list.

```css
@st .gallery {
    @st ::navBtn => .btn;
    @st ::multi => nav > :is(.a, .b);
    @st ::compound => &.x;
    @st ::compoundMulti => &:is(.x, .y);
}
```

**Equivalent today**
```css
/* gallery.st.css */
.root {}
@custom-selector :--navBtn .btn;
@custom-selector :--multi nav > .a, .b;
@custom-selector :--compound .root.x;
@custom-selector :--compoundMulti .root.x, .root.y;
```

**Usage**
```css
.gallery::navBtn {}
.gallery::multi {}
.gallery::compound {}
.gallery::compoundMulti {}

/* transforms to*/
.ns__gallery .ns__btn {}
.ns__gallery :is(nav > .ns__a, .ns__b) {}
.ns__gallery.ns__x {}
.ns__gallery:is(.ns__x, .ns__y) {}
```

[...]

PR #2862, opened 2023-05-15
feat: experimental `@st` structure definition v0

idoros:
This PR is the first step in our plan for an improved way to define a stylesheet structure. 

Changes in this PR add the ability to define nested `pseudo-elements` in a single stylesheet and focus on new syntax for defining `classes`, `pseudo-elements`, and `pseudo-states`. The new syntax allows to define deep nested `pseudo-elements` in the same stylesheet and currently enforce limits that exist in stylable (will be loosed in the future):
- classes with extend + global mapping that doesn't effect the type
    - `@st .x :is(.y) => :global(.y)` 
- pseudo-element with mapping (no extend)
    - `@st ::x => .y`

More information on this effort can be found in the [proposal](https://github.com/wix/stylable/issues/2848).

### Refactor parts handling

Up until now `pseudo-elements` were registered directly by `css-class`/ `st-custom-selector` features.
This PR add a higher level feature `st-structure` to handle the previous auto parts mode (now called **legacy-flat** mode) and the new **structure** mode.

- `st-structure` checks the mode on analyze start and register parts according to mode
    -  **legacy-flat**: set parts on `.root` class (from analyzed `pseudo-elements` and `classes`)
    -  **structure**: analyze `@st ::part` definitions and register to the parent `class` or `pseudo-element` 
    _(nested pseudo-element is a new feature)_

### New `@st` definition

**syntax:** `@st .class`

- register top level `@st .class` to `class-symbol[-st]`
    - [x] error on non class param _(remove in v3)_
    - [x] error on definition not at top level
    - [x] collect `=>` selector mapping (allow only global mapping for now)
    - [x] error on imported definition override
- inheritance (extends) - `@st .class :is(<class-selector>)`
    - [x] register top level class extended reference to `class-symbol[-st-extends]` - allow a single class extension
    - [x] transform inherited parts and states
- opt-in to new api
    - [x] console.warn about experimental feature
    - [x] when `@st` is used `st-part` ignores legacy parts (registered from `@custom-selector` and standalone css-class)
    - [x] no stylesheet implicit root class when using the new `@st` statement
    - [x] error on `-st-extends/-st-states/-st-global` (on class defined with `@st .c<class>`

### Nested pseudo-state

**syntax:** `@st :state;`

- [x] register to `class-symbol[-st-states]`
    - transform selector
- [x] diagnostics
    - state limitations (no dash prefix, reserved natives)
    - out of context state definition
    - error on state re-declare (NEW JUST FOR `@st :state`)
- language service
    - [x] suggest nested state completion - `.class:state::part` \ `.class::part:state`

### Nested pseudo-element

**syntax:** `@st ::part => mapped-selector`

- [x] top-level class parts
- [x] deep parts - new `PartSymbol`
- [x] pseudo-state definition
- [x] accept nest-selector (`&`) in mapping (similar to `.root.x` in `@custom-selector`)
- diagnostics
    - [x] error on redefine
        - including hard-coded native APIs __(solved with `@st-override` in v2)__
    - [x] error missing selector mapping
    - [x] error on multiple mapped selectors - for now this can be achieved by wrapping the selectors with `:is()` or `&is()`
    - [x]  error out-of-context definition (directly in top level `@st .class` or `@st ::pseudo-element`)
- language service
    - [x] suggest nested parts completion - `.class::partA::partB`

## Technical debts

- decide if to set `@st .x` class as root? maybe provide a way to opt-in/out?
- check why the default formatter makes mess of `@st ::part => [mapped-selector]`
- change `css-class->st-symbol` redeclare api to handle top level class in structure mode (error on declare)
- lsp for new `@st` definition
    -  top level `@st` completion
    - class definition completion 
        - class name: `@st .` / `@st .<known-undefined-class>`
        - extend class: `@st .class :is(.base)`
        - global map: `@st .class => :global(<selector>)`
    - pseudo-element completion - `@st ::part => <selector>;`
    - pseudo-state completion - `@st :<state-def>;`
    - go to definition for new definitions


packages/core/src/features/st-custom-selector.ts
```diff
@@ -122,8 +121,14 @@ export function isScoped(meta: StylableMeta, name: string) {
 }
 
 export function getCustomSelector(meta: StylableMeta, name: string): SelectorList | undefined {
+    return plugableRecord.getUnsafe(meta.data, dataKey)[name]?.ast;
```
barak007, line 124: align type

packages/core/src/stylable-transformer.ts
```diff
@@ -644,25 +642,36 @@ function validateScopes(transformer: StylableTransformer, meta: StylableMeta) {
...
+        const matchNode = structureMode
+            ? (node: SelectorNode) => node.type === 'nesting'
+            : (node: SelectorNode) => node.type === 'class' && node.value === meta.root;
+        for (let i = 0; i < first.nodes.length; i++) {
+            const node = first.nodes[i];
+            if (node.type === 'comment') {
+                continue;
+            }
+            if (matchNode(node)) {
                 hadRoot = true;
```
barak007, line 665: change hasRoot to a better name. now it can be under nesting

barak007: we should error when there is `&` not at the start of the mapping

packages/language-service/test/lib-new/features/ls-st-structure.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+                @st ::shallow => [shallow] {
+                    @st ::deep => [deep];
+                }
+            }
+
+            .cls^afterCls^ {}
+
+            .cls::shallow^afterShallow^ {}
+
+            .root^afterRoot^ {}
```
barak007, line 17: check afterRoot

barak007: remove afterRoot from this test

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+import { expect } from 'chai';
+
+const transformerStringDiagnostics = diagnosticBankReportToStrings(transformerDiagnostics);
+const stStructureDiagnostics = diagnosticBankReportToStrings(STStructure.diagnostics);
+const stStateDiagnostics = diagnosticBankReportToStrings(STCustomState.diagnostics);
+const classDiagnostics = diagnosticBankReportToStrings(CSSClass.diagnostics);
+
+type FuncParameters<F> = F extends (...args: any[]) => any ? Parameters<F> : never;
+
+const spy = <T, N extends keyof T>(target: T, funcName: N) => {
```
barak007, line 22: unify with `createSpy`

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+    return {
+        calls,
+        restoreSpy() {
+            target[funcName] = origin;
+        },
+    };
+};
+
+describe('@st structure', () => {
+    it('should warn experimental feature', () => {
```
barak007, line 49: we don't want to warn we want to info

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+                @st ::abc;
+
+                /* @analyze-error(pseudo-class) ${stStructureDiagnostics.UNSUPPORTED_TOP_DEF()} */
+                @st :abc;
+
+                /* @analyze-error(multi classes) ${stStructureDiagnostics.UNSUPPORTED_TOP_DEF()} */
+                @st .a.b;
+            `);
+        });
+        it('should extend another class', () => {
```
barak007, line 128: change title to include compose

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+            `);
+
+            const { meta, exports } = sheets['/entry.st.css'];
+
+            shouldReportNoDiagnostics(meta);
+            expect(exports.classes.abc, 'composed inline def').to.eql(
+                'entry__abc entry__defined-class-inline'
+            );
+        });
+        it('should report expected missing extended class reference', () => {
```
barak007, line 148: disallow unsupported extends

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+        it('should report selector mapping diagnostics', () => {
+            testStylableCore({
+                'general.st.css': `
+                    /* @analyze-error(empty) ${stStructureDiagnostics.MISSING_MAPPED_SELECTOR()} */
+                    @st .empty => ;
+        
+                    /* @rule(not global) .general__empty */
+                    .empty {}
+
+                    /* @analyze-error(multi) ${stStructureDiagnostics.MULTI_MAPPED_SELECTOR()} */
```
barak007, line 193: In this error we should redirect the user to use `:is()`

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+                    .not-global {}
+
+                    /* @analyze-error(not global pseudo) ${stStructureDiagnostics.GLOBAL_MAPPING_LIMITATION()} */
+                    @st .top-level-is => :is(.xyz);
+        
+                    /* @rule(not global pseudo) .global__top-level-is */
+                    .top-level-is {}
+
+                    /* @analyze-error(multi global) ${stStructureDiagnostics.GLOBAL_MAPPING_LIMITATION()} */
+                    @st .multi-global => :global(.a, .b);
```
barak007, line 213: suggest to user to use `:global(:is())`

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+        it('should report parsing issues', () => {
+            testStylableCore(`
+                @st .x {
+                    /* @analyze-warn ${stStateDiagnostics.NO_STATE_TYPE_GIVEN('bool')}*/
+                    @st :bool();
+
+                    /* @analyze-error ${stStateDiagnostics.STATE_STARTS_WITH_HYPHEN('-dashProxy')}*/
+                    @st :-dashProxy;
+
+                    /* @analyze-warn ${stStateDiagnostics.RESERVED_NATIVE_STATE('is')}*/
```
barak007, line 264: add test to check lestovers after valid parsing

packages/core/test/features/st-structure.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+            `);
+        });
+        it('should report mapped selector issues', () => {
+            testStylableCore(`
+                @st .x {
+                    /* @analyze-error ${stStructureDiagnostics.MISSING_MAPPED_SELECTOR()}*/
+                    @st ::missing-selector => ;
+
+                    /* @analyze-error ${stStructureDiagnostics.MULTI_MAPPED_SELECTOR()}*/
+                    @st ::multi-selector => .a, .b;
```
barak007, line 421: same `:is`

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+    analyzedDefs: WeakMap<postcss.AtRule, AnalyzedStDef>;
+    analyzedDefToPartSymbol: Map<AnalyzedStDef, PartSymbol>;
+}>('st-structure');
+
+// HOOKS
+
+export const hooks = createFeature({
+    analyzeInit(context) {
+        const { meta } = context;
+        const stAtRule = meta.sourceAst.nodes.find((node) => isStAtRule(node));
```
barak007, line 111: don't do work on `css` files

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+                context.diagnostics.report(diagnostics.UNSUPPORTED_TOP_DEF(), {
+                    node: atRule,
+                });
+            } else {
+                context.diagnostics.report(diagnostics.INVALID_ST_DEF(atRule.params), {
+                    node: atRule,
+                });
+            }
+        } else if (analyzed.type === 'topLevelClass') {
+            // ToDo: error when nested (only top level for now)
```
barak007, line 149: todo

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+                });
+            }
+            const existingSymbol = STSymbol.get(context.meta, analyzed.name);
+            if (existingSymbol?._kind === 'import') {
+                context.diagnostics.report(diagnostics.OVERRIDE_IMPORTED_CLASS(), {
+                    node: atRule,
+                });
+                return;
+            }
+            // ToDo: pass atuRule for diagnostics
```
barak007, line 162: todo

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+                    context,
+                    atRule,
+                    '.' + analyzed.name,
+                    '-st-extends',
+                    getAlias(extendedSymbol) || extendedSymbol
+                );
+            }
+            // class mapping
+            if (analyzed.mappedSelector) {
+                const firstSelectorNodes = analyzed.mappedSelector.nodes;
```
barak007, line 179: handle comments before global

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+            }
+        } else if (analyzed.type === 'part') {
+            const parentSymbol =
+                analyzed.parentAnalyze.type === 'topLevelClass'
+                    ? CSSClass.get(context.meta, analyzed.parentAnalyze.name)
+                    : analyzedDefToPartSymbol.get(analyzed.parentAnalyze);
+
+            if (!parentSymbol) {
+                // assuming analyzing @st definitions dfs - class/part must be defined
+                return;
```
barak007, line 221: throw

barak007: or diagnostic

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+            }
+            // analyze mapped selector
+            analyzeRule(
+                postcss.rule({
+                    selector: stringifySelectorAst(analyzed.mappedSelector),
+                    source: atRule.source,
+                }),
+                {
+                    isScoped: false,
```
barak007, line 240: should be determine from the parent chain. (global mapping)

packages/core/src/features/st-structure.ts
```diff
@@ -0,0 +1,564 @@
...
+            // register part mapping to parent definition
+            const partSymbol = setPart(parentSymbol, getSymbolId(parentSymbol), partName, [
+                analyzed.mappedSelector,
+            ]);
+            analyzedDefToPartSymbol.set(analyzed, partSymbol);
+        } else if (analyzed.type === 'state') {
+            const parentSymbol =
+                analyzed.parentAnalyze.type === 'topLevelClass'
+                    ? CSSClass.get(context.meta, analyzed.parentAnalyze.name)
+                    : analyzedDefToPartSymbol.get(analyzed.parentAnalyze);
```
barak007, line 253: extract to function

packages/core/test/features/st-structure.spec.ts
idoros: this is a `console.warn`, it doesn't break any builds, and I prefer to to add `info` diagnostics to every call to the new API

packages/core/test/features/st-structure.spec.ts
idoros: already the case

packages/core/src/features/st-structure.ts
idoros: Passed, but currently there are no actual diagnostics in the class symbol creation, so no tests.
Potentially there is the inner st-symbol re-declare diagnostic, but there refactoring the css-class registration is a bit of a sidetrack, so added it to the tech-debt to resolve after this PR.

packages/core/src/features/st-structure.ts
idoros: I believe it should just be set to `true`.

Currently the parent of a pseudo-element must be a class, that is by definition local. And in the case that a class has global mapping, by definition global mapping allow to map to global (unscoped), so there shouldn't be an error on the top level class or nested parts.

barak007 (approved):

PR #2864, opened 2023-05-30
fix: add sorting to manifest plugin to avoid content hash changes

barak007:


idoros (changes requested):
should move to v5

idoros:
closed in favor of: https://github.com/wix/stylable/pull/2877

PR #2867, opened 2023-06-08
fix(esbuild): correct resolveDir

barak007:
fixes assets resolution


packages/esbuild/src/stylable-esbuild-plugin.ts
```diff
@@ -532,7 +532,7 @@ function sortMarkersByDepth(
     const sorted = sortModulesByDepth(
         extracted,
         (m) => m.depth,
-        (m) => m.path,
+        (_) => '', // we don't want to sort by path this way we get similar results as the js injection mode
```
idoros, line 535: sort or not sort, lets try and align CLI and other integrations, runtime/static to produce the same deterministic order

idoros (approved):

PR #2868, opened 2023-06-08
fix: edit parser unclosed at-rule bug

idoros:
This PR fixes an internal bug with the `ast-from-position` for a case of unclosed at-rule at eof. Before this fix postcss would simply allow the unclosed at-rule without setting the `source.end` of the root node which caused the `ast-from-position` to ignore the unclosed at-rule.

This PR fixes:
- close all unclosed nodes at the end of safe parsing
- check for unclosed at-rule in `ast-from-position` 
- set minimum of one after the at-rule name as the params start value for the case where the space after the name is not entered yet: `@name|`

### DIFF
--- a/packages/language-service/src/lib-new/ast-from-position.ts
+++ b/packages/language-service/src/lib-new/ast-from-position.ts
@@ -74,7 +74,7 @@ export interface AstLocationResult {
 }
 function isClosed(node: postcss.AnyNode) {
     const isLast = node.parent && node.parent.nodes[node.parent.nodes.length - 1] === node;
-    if (node.type === 'decl') {
+    if (node.type === 'decl' || (node.type === 'atrule' && !node.nodes)) {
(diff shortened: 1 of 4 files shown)


packages/language-service/src/lib-new/ast-from-position.ts
```diff
@@ -269,7 +269,7 @@ function checkDeclValue(node: postcss.AnyNode, checkContext: CheckContext) {
 function checkAtRuleParams(node: postcss.AnyNode, checkContext: CheckContext) {
     if (isAtRule(node)) {
         const valueStart =
-            checkContext.baseNodeOffset + 1 + node.name.length + node.raws.afterName!.length;
+            checkContext.baseNodeOffset + 1 + node.name.length + (node.raws.afterName!.length || 1);
```
barak007, line 272: use `Math.max` ?

barak007 (approved):

PR #2870, opened 2023-06-08
fix(esbuild): cache layer for the transform process

barak007:
Since esbuild has no cache we need to manually implement it for the plugin.
 
This PR implement cache (naively check any deep dependency) for the transformation of stylable files.

more to caching should be done to other hooks.

### DIFF
--- a/packages/esbuild/src/build-cache.ts
+++ b/packages/esbuild/src/build-cache.ts
@@ -1,0 +1,55 @@
+import { statSync } from 'fs';
+import type { OnLoadResult } from 'esbuild';
+
+// mtimes of known files shared between builds
+const fileInfo = new Map<string, number>();
+
+export function buildCache() {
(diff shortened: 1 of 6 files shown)


packages/esbuild/src/build-cache.ts
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return undefined;
+    };
+    const addToCache = (path: string, result: OnLoadResult) => {
+        cache.set(path, result);
+        for (const watchFile of result.watchFiles || []) {
+            if (fileInfo.has(watchFile)) {
+                continue;
+            }
+            fileInfo.set(watchFile, statSync(watchFile).mtimeMs);
```
idoros, line 43: did you check that stylable breaks the build in case a dependency doesn't exist in a way that won't reach this point?

PR #2871, opened 2023-06-08
fix: collection if mixin declaration under rule block + more

barak007:


packages/core/test/features/st-mixin.spec.ts
```diff
@@ -41,6 +41,62 @@ describe(`features/st-mixin`, () => {
...
+                    after: 2;
+                }
+            }
+        `);
+
+        const { meta } = sheets['/entry.st.css'];
+
+        shouldReportNoDiagnostics(meta);
+    });
+    it.skip(`should keep other nested rules`, () => {
```
barak007, line 70: we need to make the testkit support nesting.

idoros: What do you mean? - you can place `@rule <selector>` above each of the nested rules?

idoros: I see now, the top checks both declarations and rules,  but the test-kit doesn't support it.

idoros (approved):

PR #2874, opened 2023-06-12
feat(lsp): filter completions by context

idoros:
This PR adds a way to allow stylable lsp plugins to check context before completion in order to take ownership of context and filter out other providers.

Currently implemented to filter out native-css service and selector completions from `@st-import` at-rule 

In addition a small refactor to change "provide" that would only be for completion into "lsp plugin" to group general lsp api around a feature/concept

### DIFF
--- a/packages/language-service/src/lib-new/lang-service-context.ts
+++ b/packages/language-service/src/lib-new/lang-service-context.ts
@@ -33,7 +33,7 @@ export class LangServiceContext {
     public ambiguousNodes: Map<any, ParseReport[]>;
     public location: ReturnType<typeof getAstNodeAt>;
     public document: TextDocument;
-    private flags = {
+    public flags = {
         runNativeCSSService: true,
     };
     constructor(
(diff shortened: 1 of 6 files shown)


packages/language-service/src/lib-new/lang-service-context.ts
```diff
@@ -56,6 +59,12 @@ export class LangServiceContext {
             this.fileData.content
         );
     }
+    public setFlag(flag: keyof LangServiceContext['flags'], enabled: boolean) {
```
barak007, line 62: are we in java? I prefer not to use strings and just make the flags public.

packages/language-service/src/lib/provider.ts
```diff
@@ -85,8 +85,8 @@ function findLast<T>(
 }
 
 export class Provider {
-    private providers: CompletionProvider[] = [
-        newStImportCompletionProvider,
+    private plugins: LangServicePlugin[] = [
+        StImportPlugin,
         RulesetInternalDirectivesProvider,
```
barak007, line 90: rename all `Provider -> Plugin`?

packages/language-service/src/lib-new/lang-service-context.ts
idoros: I felt Java and shame while writing this...

tomrav (approved):

PR #2880, opened 2023-06-28
feat: added a way to force states on a single selector

barak007:
Component libraries that uses this plugin and apis, used to copy it and change it to allow this functionality. they also made it worse by creating the context for each transformation.

now it's possible to share the context and transform only the selector ast

```ts
const context = createForceStatesContext(dataPrefix, namespaceMapping); // create once
mutateWithForceStates(parseCssSelector('.x:hover'), context); // use in the process
```

* added js docs to public api

### DIFF
--- a/packages/webpack-extensions/src/stylable-forcestates-plugin.ts
+++ b/packages/webpack-extensions/src/stylable-forcestates-plugin.ts
@@ -37,7 +37,7 @@ export function createDataAttr(dataAttrPrefix: string, stateName: string, param?
     return `${dataAttrPrefix}${paramWithValueExtraDil}${stateName}${statePart}`;
 }
 /**
- * 
+ *
  * @param targetAst - target ast to transform
  * @param namespaceMapping - known namespaces (use to detect framework specific namespaces like Stylable)
  * @param dataPrefix - prefix for the data attribute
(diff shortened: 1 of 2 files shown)


packages/webpack-extensions/src/stylable-forcestates-plugin.ts
```diff
@@ -93,22 +64,91 @@ export interface AddForceStateSelectorsContext {
...
+    );
+}
+/**
+ * creates a context to share across the transformation of addForceStateSelectors
+ * @param dataPrefix - prefix for the data attribute
+ * @param namespaceMapping - known namespaces (use to detect framework specific namespaces like Stylable)
+ * @param mapping - mapping of the original state to the override state
+ * @param plugin - plugin to override the context
+ */
+export function createContext(
```
idoros, line 87: createForceStateContext

packages/webpack-extensions/src/stylable-forcestates-plugin.ts
```diff
@@ -93,22 +64,91 @@ export interface AddForceStateSelectorsContext {
...
 
-export function addForceStateSelectors(ast: postcss.Root, context: AddForceStateSelectorsContext) {
-    ast.walkRules((rule) => {
-        const selectorAst = parseCssSelector(rule.selector);
+/**
+ * adds force state selectors to the entire ast
+ * @param targetAst - target ast to transform
+ * @param context - context to use for the transformation
+ */
+export function addForceStateSelectors(
```
idoros, line 72: export from index

idoros (approved):

PR #2882, opened 2023-07-03
fix: support svg `p: path("")` declaration

idoros:
This PR fixes the ability to use `d: path()` to [control the path](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#moveto_path_commands) of an svg path element:

- preserve quotes of `path("")` argument (invalid without)
- suppress native-css-service `Unknown property: 'd'`
- check for native "path" only after custom JS formatters to allow custom override (would be a breaking check if not)


tomrav (approved):

packages/core/src/native-reserved-lists.ts
```diff
@@ -102,86 +102,81 @@ export const nativePseudoElements = [
...
-    swash: true,
-    symbols: true,
-    translate: true,
-    translate3d: true,
-    translateX: true,
-    translateY: true,
-    translateZ: true,
-    url: true,
-    var: true,
+export const nativeFunctionsDic: Record<string, { preserveQuotes: boolean }> = {
```
barak007, line 105: is it public? I would keep another dict with this info. no need to change this record.

barak007: This should have been a Set anyway

idoros: Not public, used only in this specific one place. 

I would change the behavior to always preserve whatever was in the source for native value functions, but that is a bigger change that I just don't have the time to check for atm.

We need to organize all the "native/spec" at some point and also refactor the "functions" module.

PR #2883, opened 2023-07-04
feat: st-scope with no scoping selector

idoros:
This PR relaxes the restrictions on the usage of `@st-scope`, allowing it to be used solely for opting out of global side-effect restrictions. This means that there is no longer a need to prepend a scoping selector to any nested selectors.

The @st-scope directive offers two capabilities:

- It allows for opting out of `unscoped` selector diagnostics, which is useful for defining a theme or resetting with global side-effects.
- It reduces the repetition of selectors.

With native CSS nesting support, the second capability becomes redundant.

To address the opt-out feature more clearly, we can consider the following approaches:

1. Use eslint type instructions, such as `@st-ignore "unscoped"`.
2. Explicitly state the semantics of the stylesheet using statements like `@st-global, @st-theme, or @st-side-effects`.

For now, allowing no selector allows a similar type of behavior.

I didn't report a new error for invalid selectors or comments, as this would be a breaking change.

### DIFF
--- a/packages/core/src/features/st-scope.ts
+++ b/packages/core/src/features/st-scope.ts
@@ -20,38 +20,18 @@ export const hooks = createFeature<{ IMMUTABLE_SELECTOR: ImmutablePseudoClass }>
         if (!isStScopeStatement(atRule)) {
             return;
         }
-        const selectorAst = parseSelectorWithCache(atRule.params);
-        const isValidSelector =
-            selectorAst.length !== 0 &&
-            selectorAst.every((selector) => {
-                let foundSelector = false;
-                for (const { type } of selector.nodes) {
-                    switch (type) {
-                        case 'invalid':
-                            return false;
-                        case 'comment':
-                            break;
-                        default:
-                            foundSelector = true;
-                    }
-                }
-                return foundSelector;
-            });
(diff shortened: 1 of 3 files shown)


packages/core/src/features/st-scope.ts
```diff
@@ -21,19 +20,38 @@ export const hooks = createFeature<{ IMMUTABLE_SELECTOR: ImmutablePseudoClass }>
         if (!isStScopeStatement(atRule)) {
             return;
         }
-        if (!atRule.params) {
-            context.diagnostics.report(diagnostics.MISSING_SCOPING_PARAM(), { node: atRule });
+        const selectorAst = parseSelectorWithCache(atRule.params);
+        const isValidSelector =
```
barak007, line 24: isValidSelector? I'm not sure what do you test here

idoros: that the params, when parsed as a selector list:
- have at least 1 selector
- each selector doesn't have an invalid node
- each selector have some actual selector (not just comments)

idoros: removed the check and allowed any params to be analyzed as a selector.

barak007 (approved):

PR #2887, opened 2023-07-06
fix(lsp): unhandled errors

idoros:
This PR wraps the main language service APIs, catch and log any unexpected errors, and return a fallback default value.

It should minimize the errors a user sees, and still allow us to research issues.

-----

In addition there is a small fix for `onSignatureHelp` that explodes:

```
.x {
    -st-mixin: mix(
        arg value(var)^caret:press(ctrl+shift+space)^
    )
}
```

### DIFF
--- a/packages/language-service/src/lib/utils/wrap-and-log.ts
+++ b/packages/language-service/src/lib/utils/wrap-and-log.ts
@@ -1,6 +1,6 @@
 type FunctionConfig<T> = Pick<
     {
-        [K in keyof T]: T[K] extends (...args: any[]) => infer R ? R : never;
+        [K in keyof T]: T[K] extends (...args: any[]) => infer R ? () => R : never;
     },
     {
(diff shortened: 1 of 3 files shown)


packages/language-service/src/lib/utils/wrap-and-log.ts
```diff
@@ -0,0 +1,34 @@
...
+        const func = proto[name];
+        if (typeof func !== 'function') {
+            console.error(
+                `expected to find a function named ${name} on context, but found ${typeof func}`
+            );
+            continue;
+        }
+        proto[name as keyof T] = function (this: T, ...args: any[]) {
+            try {
+                return func.apply(this, args) || defaultReturn;
```
barak007, line 25: the return value should be create each time. please convert to a function

idoros: I believe the result is serialized, but I'll recreate it just to be safe.

PR #2890, opened 2023-07-12
Update native-reserved-list value functions

barak007:
Added `scroll` function from scroll animation timeline


idoros:
this one, right? 
https://drafts.csswg.org/scroll-animations/#scroll-notation

idoros (approved):

idoros:
from the spes we are missing the following list of value functions:

```
[
    // css-anchor-position
    anchor,
    // css-conditional-5
    supports,
    // css-color-5
    'color-mix',
    lab,
    oklab,
    lch,
    oklch,
    color,
    'device-cmyk',
    // css-color-6
    'contrast-color',
    wcag2,
    // css-conditional-4
    selector,
    // css-conditional-5
    'font-tech',
    'font-format',
    media,
    // css-contain-3
    style,
    // css-content
    'target-counter',
    'target-counters',
    'target-text',
    string,
    content,
    // css-gcpm
    running,
    // css-images-4
    'image-set',
    type,
    'cross-fade',
    element,
    stripes,
    // css-link-params
    param,
    // css-lists
    reversed,
    // css-shapes-2
    shape,
    // css-text-4
    auto,
    // css-values-5
    random,
    'random-item',
    // css-values
    mix,
    src,
    round,
    mod,
    rem,
    sin,
    cos,
    tan,
    asin,
    acos,
    atan,
    atan2,
    pow,
    sqrt,
    hypot,
    log,
    exp,
    abs,
    sign,
    // fill-stroke
    child,
    // filter-effects
    filter,
    // mathml-core
    add,
    // motion
    ray,
    // scroll-animations
    view,
]
```

Our list is not used for lsp completions, just for diagnosing unknown functions, so we might want to add them even if they aren't fullly supported yet.

PR #2896, opened 2023-08-27
fix(core): imported symbols runtime leak

idoros:
This pull request addresses a longstanding regression related to the unintended runtime leakage of imported symbols.

During the process of refactoring the internal symbol resolution code to achieve consistent handling of various symbols, an oversight occurred. This oversight caused the generated JavaScript runtime and the  `.d.ts` files to inadvertently expose imported symbols of the stylesheet.

This situation is problematic because Stylable's assumption is that symbols of a stylesheet are only meant to be used if the stylesheet is imported into a JavaScript file. If it's not imported this way, it's assumed that the symbols' namespacing can't be connected to the DOM and can thus be removed as unnecessary code from the final output.

This fix is undoubtedly necessary. However, it's important to acknowledge that there's a possibility it could introduce disruptions in external code. To mitigate this potential impact, we're planning to conduct a thorough search across both public code repositories and internal code projects accessible to us. This search aims to identify instances where this fix might cause conflicts or complications. Based on our findings, we will make an informed decision about whether to integrate this fix into the upcoming v5 release or reserve it for inclusion in the subsequent major release, v6.

### DIFF
--- a/packages/module-utils/test/generate-dts.spec.ts
+++ b/packages/module-utils/test/generate-dts.spec.ts
@@ -221,10 +221,13 @@ describe('Generate DTS', function () {
             'origin.st.css': `
                 .cls {}
                 @property --customProp;
+                :vars {
+                    buildVar: red;
+                }
             `,
             'test.st.css': `
-                @st-import CompRoot, [cls, --customProp] from './origin.st.css';
-                .cls { color: green; }
+                @st-import CompRoot, [cls, --customProp, buildVar] from './origin.st.css';
+                .cls { color: value(buildVar); }
                 .CompRoot { 
                     color: green;
                     --customProp: green;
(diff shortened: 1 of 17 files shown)


packages/module-utils/test/generate-dts.spec.ts
```diff
@@ -215,6 +216,61 @@ describe('Generate DTS', function () {
...
+                eq<string>(vars.customProp);
+                eq<string>(vars.customProp);
+                eq<string>(stVars.buildVar);
+                eq<string>(keyframes.anim);
+                eq<string>(containers.cont);
+                eq<string>(layers.layer1);
+            `,
+        });
+
+        expect(tk.typecheck('test.ts')).to.include(propNotOnType('cls'));
```
barak007, line 265: How does this does checks that something is not exported?

packages/module-utils/test/sourcemap.spec.ts
```diff
@@ -163,14 +163,9 @@ describe('.d.ts source-maps', () => {
         const res = generateStylableResult({
             entry: `/entry.st.css`,
             files: {
-                '/another.st.css': {
```
barak007, line 166: so the entire behavior is expected?

packages/module-utils/test/generate-dts.spec.ts
idoros: The test import symbols from `origin` stylesheet to `test` stylesheet and then checks that d.ts that describes the runtime doesn't contain any leaked imported symbols.

The actual runtime maps are checked as part of the core. I might have skipped over testing the runtime e2e - will add!

packages/module-utils/test/sourcemap.spec.ts
idoros: not sure what you mean

barak007 (approved):

PR #2898, opened 2023-09-06
fix: mixin with native CSS nesting

idoros:
This PR fixes some cases of mixins with native CSS nesting:

- mixin with nesting rules
- mix into a nested rule
- preserve nesting selector inside mixed fragment
- error diagnostic in case of a mixin class that is found inside a nested fragment (temporary until we figure out how to handle this case better)
- mix nested at-rule with content
- preserve nesting rules and declaration validity when mixed in an invalid order (nest declarations following nested rules with `&` selector)

### DIFF
--- a/packages/core/test/features/st-mixin.spec.ts
+++ b/packages/core/test/features/st-mixin.spec.ts
@@ -6,9 +6,6 @@ import {
     shouldReportNoDiagnostics,
     matchRuleAndDeclaration,
     diagnosticBankReportToStrings,
-    assertRule,
-    assertDecl,
-    assertAtRule,
     deindent,
 } from '@stylable/core-test-kit';
 import chai, { expect } from 'chai';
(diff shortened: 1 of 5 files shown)


packages/core/test/features/st-mixin.spec.ts
```diff
@@ -409,37 +406,174 @@ describe(`features/st-mixin`, () => {
...
+                /* @transform-error ${mixinDiagnostics.INVALID_RECURSIVE_MIXIN()}*/
+                .entry__into {
+                    &&.entry__inner {
+                        color: green;
+                    }
+                }
+            `)
+            );
+        });
+        it('should mix nested at-rules with content', () => {
```
barak007, line 540: we should test the case when there are decls after the -st-mixin.

idoros: check for what? that we don't mix nested rules before declarations? split the rule in that case?

barak007: either that or not allowing

idoros: went with nesting in case of declaration and nested rules out of order

barak007 (approved):

PR #2906, opened 2023-09-20
feat!: turn on experimental selector inference by default

idoros:
This PR flips the default, so that the `experimentalSelectorInference` flag is on unless explicitly set to `false` -  the false behavior is kept for now, but will be removed in the future within version 6.

- [x]  set default to true
- [x] report false as unsupported mode that will be removed in a future version
- [x] change test expectations that relay on the old behavior
- [x] refactored tests to match new default (kept some experimentalSelectorInference=false to test previous behavior)

> As part of the upgrade in the vscode extension, a task must be undertaken to verify the installed `@stylable/core` version and, if not configured by the user, override the default `experimentalSelectorInference` setting according to the detected version (false for <=5, true for >=6).


packages/core/test/features/st-custom-selector.spec.ts
```diff
@@ -59,17 +60,15 @@ describe('features/st-custom-selector', () => {
...
         testStylableCore(`
             /* @analyze-error(in custom) word(:--unknown) ${customSelectorDiagnostics.UNKNOWN_CUSTOM_SELECTOR(
                 ':--unknown'
             )} */
             @custom-selector :--x .before:--unknown.after;
 
             /* 
-                @transform-error(in selector) word(:--unknown)  ${customSelectorDiagnostics.UNKNOWN_CUSTOM_SELECTOR(
-                    ':--unknown'
+                @transform-error(in selector) word(--unknown)  ${pseudoClassDiagnostics.UNKNOWN_STATE_USAGE(
```
barak007, line 70: what's behind this change?

barak007 (approved):

packages/core/test/features/st-custom-selector.spec.ts
idoros: before the switch, inline custom selectors were transformed away before transformation process, and failed early, now they are transformed inline the transformation process, so they don't fail on unknown custom selector, but on unknown pseudo-state. 

And although you can't currently create a custom state with an intial `--`, I would like to add it in a future change.

PR #2911, opened 2023-10-01
chore!: deprecate resolveExtends of custom-selector

idoros:
This PR removes the optional `transformer` argument from `resolver.resolveExtends` function.

Passing the transformer argument currently instructs the resolver to first seek a custom selector before inspecting the stylesheet meta for a class and then determining its type accordingly.

There are several reasons for removing this API:
- It is not utilized within Stylable or, to the best of our knowledge, outside of the Stylable repository.
- This functionality relies on custom-selector, which we intend to deprecate in favor of the new structure-mode with mapped selectors, which provides an alternative method for deducing the extended type.
- This is the only workflow that necessitates the presence of the transformer argument in the resolver, complicating our refactoring plans.

In the future, there may be a need for a similar method to resolve such issues, and we anticipate reintroducing similar functionality when it becomes necessary.

### DIFF
--- a/packages/core/src/stylable-resolver.ts
+++ b/packages/core/src/stylable-resolver.ts
@@ -9,7 +9,6 @@ import {
     StylableSymbol,
     CSSClass,
     STSymbol,
-    STCustomSelector,
     VarSymbol,
     CSSVarSymbol,
     KeyframesSymbol,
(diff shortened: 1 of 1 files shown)


packages/core/src/stylable-resolver.ts
```diff
@@ -440,23 +436,6 @@ export class StylableResolver {
                     : meta.getClass(nameOrSymbol)
                 : nameOrSymbol;
 
-        const customSelector = isElement
```
barak007, line 443: maybe we should throw if we encounter this?

idoros: We have no internal flows that assumes this, so throwing without the `transformer` flag will probably break in case you have a custom-selector override in an unexpected manner.

barak007 (approved):

PR #2912, opened 2023-10-01
refactor(types)!: enhance type definitions for resolved CSS and JS values

idoros:
This PR enhances the type definitions for resolved CSS and JS values in the following ways:

1. The CSS resolved type now accurately represents the scenario where the `symbol` field can be `undefined` when the stylesheet exists, but the queried symbol definition is missing.
2. The JS resolved symbol type (exported value) has been updated from the loose `any` type to the more precise `unknown` type.

### DIFF
--- a/packages/core/src/functions.ts
+++ b/packages/core/src/functions.ts
@@ -187,7 +187,9 @@ export function processDeclarationValue(
                 const formatterArgs = getFormatterArgs(parsedNode);
                 try {
                     // ToDo: check if function instead of calling on a non function
-                    parsedNode.resolvedValue = (formatter.symbol as Function)(...formatterArgs);
+                    parsedNode.resolvedValue = (formatter.symbol as (...args: any[]) => any)(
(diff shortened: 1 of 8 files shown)


packages/core/src/functions.ts
```diff
@@ -186,7 +186,8 @@ export function processDeclarationValue(
                 const formatter = resolvedSymbols.js[value];
                 const formatterArgs = getFormatterArgs(parsedNode);
                 try {
-                    parsedNode.resolvedValue = formatter.symbol.apply(null, formatterArgs);
+                    // ToDo: check if function instead of calling on a non function
+                    parsedNode.resolvedValue = (formatter.symbol as Function)(...formatterArgs);
```
barak007, line 190: as Function

idoros: changed to `(...args: any[]) => any`

barak007 (approved):

PR #2919, opened 2023-10-16
chore: upgrade to support for `rollup@4`

idoros:
This PR introduces an upgrade to the tested rollup integration, bumping it up to [version 4](https://github.com/rollup/rollup/releases/tag/v4.0.0), and ensure it works.

During testing, one issue surfaced where [files loaded via the integration's load hook were not automatically watched](https://github.com/rollup/rollup/pull/5150). This update addresses this issue by explicitly adding them to the list of watched files within the integration.

Furthermore, it's worth noting that `rollup@4` installation includes WASM artifacts, which, in some cases, led to failed installation on Windows machines. This problem was resolved by ensuring that Microsoft Visual C++ is installed. The installer for Microsoft Visual C++ can be obtained from this link: https://aka.ms/vs/17/release/vc_redist.x64.exe.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "rimraf": "^5.0.5",
-        "rollup": "^3.29.4",
+        "rollup": "^4.1.4",
         "source-map": "^0.7.4",
         "source-map-loader": "^4.0.1",
         "style-loader": "^3.3.3",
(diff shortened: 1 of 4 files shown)


barak007:
> Another important change is that Rollup watch mode will no longer watch ids of files that have been loaded via a plugin [load](https://rollupjs.org/plugin-development/#load) hook. So this mainly affects "virtual" files, where it really does not make sense to watch a hard drive location for changes. Instead, it is now up to plugins that use a load hook to manually call [this.addWatchFile()](https://rollupjs.org/plugin-development/#this-addwatchfile) for all the files they depend on to handle the load hook.

Did you checked all locations?

idoros:
The only change is the load hook

barak007 (approved):

PR #2925, opened 2023-11-20
fix(core): scope nested global selector issues

idoros:
This pr fix the code that combines 2 selectors into a single selector in order to fix 2 bugs:

- when nesting a global selector into a `@st-scope`, the   parent selector is ignored
- when mixining a JS mixin, any nested rule that starts with `:global()` is missing the mixed into selector

### DIFF
--- a/packages/core/src/helpers/selector.ts
+++ b/packages/core/src/helpers/selector.ts
@@ -199,7 +199,8 @@ export function scopeNestedSelector(
             );
             // merge scope flags
             const nestStartWithNesting = first.type === `nesting`;
-            const nestedStartWithGlobal = first.type === `pseudo_class` && first.value === `global`;
+            const nestedStartWithGlobal =
+                rootScopeLevel && first.type === `pseudo_class` && first.value === `global`;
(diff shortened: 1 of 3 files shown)


packages/core/src/helpers/selector.ts
```diff
@@ -197,17 +197,20 @@ export function scopeNestedSelector(
...
-            const startWithScoping = rootScopeLevel
-                ? scopeAst.nodes.every((node, i) => {
-                      return matchTypeAndValue(node, outputAst.nodes[i]);
-                  })
-                : false;
-            let nestedMixRoot = false;
+            // merge scope flags
+            const nestStartWithNesting = first.type === `nesting`;
+            const nestedStartWithGlobal =
+                rootScopeLevel && first.type === `pseudo_class` && first.value === `global`;
```
idoros, line 203: the change is here - the addition of `rootScopeLevel &&` to limit the behavior of not prepending the parent scope for root mixins

barak007 (approved):

PR #2928, opened 2023-11-26
feat: improve namespace readability and potential optimization conflict 

idoros:
This PR introduces a few improvements to the way stylesheet meta namespaces are resolved:

1. To mitigate potential conflicts involving the underscore character (_), which may clash with the target local delimiter, this update reduce multiple consecutive underscores into a single underscore.

2. Extends support for the hyphen character (-). It similarly trims consecutive hyphens to a single instance, thereby preventing issues related to state transformation conflicts. This change is particularly beneficial for handling default namespaces derived from file paths with dashes. By preserving these dashes, the resulting namespaces become more readable and accurately reflect their source file paths.

3. Extends support for non ASCII characters in the namespace value.

4. Align `@st-namespace` validation with the allowed transformed value that is generated from a stylesheet filename.

### DIFF
--- a/packages/core/src/helpers/string.ts
+++ b/packages/core/src/helpers/string.ts
@@ -24,9 +24,7 @@ export function string2varname(str: string) {
             .replace(/--+/gm, '-')
             // replace multiple underscores with single underscore
             .replace(/__+/gm, '_')
-            // remove leading if not letters, dashes, underscores or non-ascii
-            .replace(/[\x00-\x7F]+/gm, (matchAscii) => {
-                return matchAscii.replace(/^[^a-zA-Z_-]+/gm, '');
-            })
+            // remove leading digits from start
+            .replace(/^\d+/gm, '')
     );
 }
--- a/packages/core/test/features/st-namespace.spec.ts
+++ b/packages/core/test/features/st-namespace.spec.ts
@@ -21,7 +21,8 @@ describe('features/st-namespace', () => {
             '/_underscore.st.css': ``,
             '/--multi---dash.st.css': ``,
             '/__multi___underscore.st.css': ``,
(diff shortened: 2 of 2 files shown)


packages/core/src/helpers/string.ts
```diff
@@ -1,11 +1,32 @@
...
+            // allow only letters, numbers, dashes, underscores, and non-ascii
+            .replace(/[\x00-\x7F]+/gm, (matchAscii) => {
+                return matchAscii.replace(/[^0-9a-zA-Z_-]/gm, '');
+            })
+            // replace multiple dashes with single dash
+            .replace(/--+/gm, '-')
+            // replace multiple underscores with single underscore
+            .replace(/__+/gm, '_')
+            // remove leading if not letters, dashes, underscores or non-ascii
+            .replace(/[\x00-\x7F]+/gm, (matchAscii) => {
```
barak007, line 28:
```suggestion
            .replace(/^\d+/gm, '')
```

packages/core/test/features/st-namespace.spec.ts
```diff
@@ -16,14 +16,31 @@ describe('features/st-namespace', () => {
     it('should use filename as default namespace', () => {
         const { sheets } = testStylableCore({
             '/a.st.css': ``,
-            '/b.st.css': ``,
+            '/B.st.css': ``,
+            '/-dash.st.css': ``,
+            '/_underscore.st.css': ``,
+            '/--multi---dash.st.css': ``,
+            '/__multi___underscore.st.css': ``,
+            '/🤡emoji🤷‍♀️.st.css': ``,
```
barak007, line 24: add one more case with [ascii][emoji][ascii]

PR #2929, opened 2023-12-03
feat: strict custom property 

idoros:
This pr introduces a new feature: the `strictCustomProperty` flag, which ensures that custom property usage always relays on a defined property and does not define a property by itself.

Previously, the behavior was more lenient, allowing custom property definition via get/set methods within a stylesheet. However, with the `strictCustomProperty` flag activated, any reference to an undeclared property (e.g. `--undeclared: red` or `color: var(--undeclared)`) will result in a diagnostic error being reported.

It's important to note that this behavior applies only to `st.css` stylesheets and not native CSS stylesheets..

Additionally, this pr introduces a new flags option in the Stylable configuration, allowing for the activation of the `strictCustomProperty` flag.

### DIFF
--- a/packages/core/src/features/css-custom-property.ts
+++ b/packages/core/src/features/css-custom-property.ts
@@ -285,7 +285,7 @@ function addCSSProperty({
             // already defined
             return;
             // eslint-disable-next-line no-constant-condition
-        } else if (context.meta.type === 'stylable' && context.flags.strictCustomProperty) {
+        } else if (context.meta.type === 'stylable' && context.meta.flags.strictCustomProperty) {
             // strict mode
             context.diagnostics.report(diagnostics.UNDEFINED_CSS_CUSTOM_PROP(name), {
                 node,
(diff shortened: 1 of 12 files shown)


packages/core/src/features/css-custom-property.ts
```diff
@@ -273,8 +278,23 @@ function addCSSProperty({
...
+    if (!final) {
+        const existing = STSymbol.get(context.meta, name, `cssVar`);
+        if (existing) {
+            // already defined
+            return;
+            // eslint-disable-next-line no-constant-condition
+        } else if (
+            context.meta.type === 'stylable' &&
+            context.flags.strictCustomProperty &&
+            !context.meta.source.includes('node_modules')
```
barak007, line 290: this is risky. maybe the flag can carry this?

packages/core/src/helpers/custom-state.ts
```diff
@@ -763,6 +764,7 @@ export const systemValidators: Record<string, StateParamType> = {
 };
 
 export function validateRuleStateDefinition(
+    flags: FeatureFlags,
```
barak007, line 767: is this a future thingy?

packages/core/src/stylable-meta.ts
```diff
@@ -61,9 +62,13 @@ export class StylableMeta {
     // Generated during transform
     public targetAst?: postcss.Root;
     public globals: Record<string, boolean> = {};
-    constructor(public sourceAst: postcss.Root, public diagnostics: Diagnostics) {
+    constructor(
+        public sourceAst: postcss.Root,
+        public diagnostics: Diagnostics,
+        flags: FeatureFlags
```
barak007, line 68: make it a member. this way we can debug it when we see the meta. also I think per meta flags

PR #2930, opened 2023-12-13
chore!: convert eslint plugin to esm

barak007:
- upgrade eslint dependencies to version 6 (esm only)
- change `eslint-plugin-stylable` to esm only to handle upgrade

more and more dependencies are moving into esm so we need to adopt.


idoros:
WE can open the discussion about moving to esm, can we support dual mode easily?

idoros (approved):

PR #2931, opened 2023-12-14
refactor: 10x better types for test kit.

barak007:
This PR making the test kit much easier to work with, sheets are auto-completed from keys or entries.

### DIFF
--- a/packages/core-test-kit/test/test-stylable-core.spec.ts
+++ b/packages/core-test-kit/test/test-stylable-core.spec.ts
@@ -71,7 +71,7 @@ describe(`testStylableCore()`, () => {
                 }
             );
 
-            const entryNotProcessed = sheets[`/entry.st.css`];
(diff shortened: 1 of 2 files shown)


packages/core-test-kit/test/test-stylable-core.spec.ts
```diff
@@ -70,7 +70,8 @@ describe(`testStylableCore()`, () => {
                     entries: [`/a.st.css`, `/b.st.css`],
                 }
             );
-
+            
+            // @ts-expect-error entry is not processed
```
idoros, line 74: why is this needed? isn't the `Keys & Entries` used for `Sheets` intersect both entries and the passed files?
If it doesn't then I would ignore the entries option and use just the files, they should include all the entries anyway.

idoros (approved):
Very nice. I would go with 2x better, but sure :)

PR #2938, opened 2024-02-08
fix: esbuild plugin break after stylable error

barak007:
in watch mode when there was transform error stylable esbuild plugin did not recover well this PR fixes that.


idoros (approved):
It would be much better to know when the build finish and not sleep fo 6+ seconds in the test

barak007:
done

PR #2942, opened 2024-03-03
feat: expose computed build vars source location

idoros:
This PR exposes the source location (meta, start, end) of computed build vars calculated using `stylable.stVar.flatten(meta)` or `stylable.stVar.getComputed(meta)`.


packages/core/src/features/st-var.ts
```diff
@@ -203,6 +209,11 @@ export function get(meta: StylableMeta, name: string): VarSymbol | undefined {
 
 // Stylable StVar Public APIs
 
+const UNKNOWN_LOCATION = {
```
barak007, line 212: freeze of spread

barak007 (approved):

PR #2951, opened 2024-03-24
fix: add missing deps and devDeps

AviVahl:
code previously got those from 4th+ parties


packages/language-service/package.json
```diff
@@ -26,6 +26,9 @@
     "vscode-languageserver-textdocument": "^1.0.11",
     "vscode-uri": "^3.0.8"
   },
+  "devDependencies": {
+    "vscode-languageserver-protocol": "^3.17.5"
```
AviVahl, line 30: this package is really specific to the language-service, so put it in that package's package.json

idoros (approved):

PR #2972, opened 2024-08-28
refactor(typescript): use node16 moduleResolution

AviVahl:
added missing project references
worked around vlq having no "types" inside its "exports"


AviVahl:
Need to validate that create-stylable-app's import of definition file works.

idoros (approved):

PR #2978, opened 2025-01-14
refactor: make imports "type": "module" ready

AviVahl:
- also renamed .mocharc.js to cjs
- should reduce the noise significantly when doing the migration


idoros (approved):
Thank you so much for this! 
Are you doing all of this manually?

AviVahl:
Mostly search/replace, but also needed to go over them one by one, reverting some changes, as non-import statements also got changed (strings in tests that import stuff).

